Issue #5, opened 2020-07-15
Roadmap to go to the v1


Nexucis:
The goal of this issue is to list the different feature we would like to implement in order to provide a nice v1.
The different issues are normally listed by priority

In order to achieve that, we are going to move directly to [CodeMirror 6](https://github.com/codemirror/codemirror.next) which provides much more interesting feature than the previous version such as:
* a grammar syntax called **[lezer](https://lezer.codemirror.net/)**
* everything is a module (or an extension as it is the internal term) which makes CodeMirror much more flexible
* CodeMirror in typescript 

## Replace Monaco
**Goal**:  be able to provide same feature than provided by [monaco-promql](https://github.com/prometheus-community/monaco-promql)

- [x] Provide the npm package **lezer-promql** that will provide the PromQL grammar in the lezer syntax. On @juliusv side
- [x] Create a language extension that takes the package **lezer-promql** in order to have a proper syntax highlightning
- [x] Create an auto-completion extension in order to suggest all PromQL keyword
- [x] In the meantime try to fix the issue https://github.com/codemirror/codemirror.next/issues/236 in order to have an accurate autocomplete list
- [x] In the meantime try to fix the issue https://github.com/codemirror/codemirror.next/issues/234

## Documentations

- [x] Document how to integrate _codemirror-mode_ with Angular
- [ ] Document how to ingegrate _codemirror-mode_ with ReactJS

## Improvement
**Goal**: improve the different feature provided in the previous section while still staying offline
- [x] Provides a way to highlight some syntax error (if not already provided) ==> use the linter. 
   - [x] Should wait the v0.10.0 to be able to display the error message: https://github.com/codemirror/codemirror.next/issues/261
   - [x] Issue https://github.com/codemirror/codemirror.next/issues/270 must be resolved to be able to iterate properly through the tree.
- [x] Improve the auto-completion list with PromQL snippets
- [x] Propose different theme to be used to replace the current one.
- [x] Hybrid Autocompletion should be deeply reviewed to be more flexible / testable / precise

## Test
- [x] Implement unit test to cover and improve potentially the autocompletion

## LSP
**Goal**: Integrate some feature coming from LSP (Language Server Protocol)

- [x] Improve the auto-completion list by getting the metric/labelName /labelValue when it's accurate. Or by using the endpoint **/completion** on Prometheus side, or by using the standard endpoint in prometheus. on @Nexucis side
[...]

PR #6, opened 2020-07-16
move to codemirror-next and propose an hightlight extension for promQL

Nexucis:
* Move to codemirror-next
* Clean the repository
* create the different file struct needed
* Propose a short hightlight extension for promQL --> needs to be improved for sure

/cc @juliusv


src/lang-promql/promql.ts
```diff
@@ -0,0 +1,28 @@
+import { LezerSyntax } from "@codemirror/next/syntax";
+import { parser } from "lezer-promql";
+import { styleTags } from "@codemirror/next/highlight";
+import { Extension } from "@codemirror/next/state";
+
+export const promQLSyntax = new LezerSyntax(parser.withProps(
+  styleTags({
+    LineComment: "comment",
+    LabelName: "name",
```
juliusv, line 9: This can be more specific, `labelName`.

src/lang-promql/promql.ts
```diff
@@ -0,0 +1,28 @@
...
+    StringLiteral: "string",
+    NumberLiteral: "number",
+    Duration: "number",
+    "abs absent absent_over_time avg_over_time ceil changes clamp_max clamp_min count_over_time days_in_month day_of_month day_of_week delta deriv exp floor histogram_quantile holt_winters hour idelta increase irate label_replace label_join ln log10 log2 max_over_time min_over_time minute month predict_linear quantile_over_time rate resets round scalar sort sort_desc sqrt stddev_over_time stdvar_over_time sum_over_time time timestamp vector year": "keyword",
+    "avg bottomk count count_values max min quantile stddev stdvar sum topk": "keyword definition",
```
juliusv, line 14: Why `keyword definition`? I just had this as `keyword` in my version.

src/lang-promql/promql.ts
```diff
@@ -0,0 +1,28 @@
...
+    "group_left group_right by without bool offset on and or unless": "operator",
```
juliusv, line 15: I split these into:

```
      'by without bool on ignoring group_left group_right offset': 'modifier',
      'and unless or': 'keyword',
```

Although the latter could also be `operator`.

src/lang-promql/promql.ts
```diff
@@ -0,0 +1,28 @@
...
+    BinOp: "operator",
+    MatchOp: "operator",
+    UnaryOp: "operator",
+    "( )": "paren",
+    "[ ]": "squareBracket",
+    "{ }": "brace",
+  })
```
juliusv, line 22: I also debugged how parse errors were tagged, and found they had funny warning sign emoji names. So to expose errors to the highlighter, I added:

```
  '⚠': 'invalid',
```

src/lang-promql/promql.ts
```diff
@@ -0,0 +1,28 @@
...
+))
```
juliusv, line 23: I also added the following, though I think it might not be needed:

```
  {
    languageData: {
      closeBrackets: { brackets: ['(', '[', '{', "'", '"', '`'] },
      commentTokens: { line: '#' },
    },
  }
```

src/lang-promql/promql.ts
Nexucis: aaah that's the purpose of this weird character. Thanks I was wondering if it was a bug or not. Thanks !

src/lang-promql/promql.ts
Nexucis: yeah I don't think it's needed.

src/lang-promql/promql.ts
Nexucis: well I can add it np, but it won't change anything normally. Since labelName is linked to the name

https://github.com/codemirror/codemirror.next/blob/master/highlight/src/highlight.ts#L166

src/lang-promql/promql.ts
Nexucis: it was a test to know if the color can switch from keyword to definition (it was coming from what I saw in the support of javascript)

src/lang-promql/promql.ts
juliusv: Also not the comment token? For automatic toggling of line comments? See https://codemirror.net/6/docs/ref/#comment

src/lang-promql/promql.ts
Nexucis: I will change it to keyword, doesn't work anyway

juliusv:
Btw., this was the theme and highlighter I used (do not include the default highlighter in combination with this):

```typescript
    const theme = EditorView.baseTheme({
      wrap: {
        // See https://github.com/codemirror/codemirror.next/blob/6ff92c76d3e57f523cc356578fd50f9457e22851/view/src/theme.ts#L50-L56
        '&.cm-focused': {
          outline: 'none',
          outline_fallback: 'none',
        },
      },
      matchingBracket: { color: 'green', backgroundColor: '#eee', outline: '1px solid #aaa' },
      nonmatchingBracket: { borderColor: 'red' },
    });

    const hl = highlighter({
      name: { color: '#000' },
      number: { color: '#09885a' },
      string: { color: '#a31515' },
      keyword: { color: '#008080' },
      labelName: { color: '#800000' },
      operator: {},
      modifier: { color: '#008080' },
      paren: {},
      squareBracket: {},
      brace: {},
      invalid: { color: 'red' },
      comment: { color: '#888', fontStyle: 'italic' },
    });
```

src/lang-promql/promql.ts
Nexucis: ah Sorry I didn't read correctly. Yes let's add it, it won't hurt if people don't want to use the closing feature

Nexucis:
Concerning the theming, if you don't mind, I would prefere to do it separately. I would prefere to propose a theme that can leave with the default highlighter like they did for the dark theme for example

src/lang-promql/promql.ts
Nexucis: I don't see the effect of adding that, but maybe I missed a configuration somewhere. In any case I added it

juliusv:
Yeah, I only provided that as an example, wouldn't integrate it tightly :)

src/lang-promql/promql.ts
Nexucis: I think `by without bool on ignoring group_left group_right offset` should have a different color than `sum avg, count`. 
That's also why I put it as operator for both of them.

What do you think ?

juliusv: Hmm, they are definitely not operators though, they are modifiers (even if `modifier` is just mapped to `keyword` in https://codemirror.net/6/docs/ref/#highlight). I would also tend to give them the same color as the aggregators (which I would classify as keywords), but maybe that's just me...

juliusv: Btw. are we sure that sub-types aren't separately styleable? Because the highlight module docs say, "Types may extend other types. When no style for a given tag is present in the highlighter, it will fall back first to styles for that type without one or more flags, and then, if that also fails, to its parent types."

Nexucis: mmm ok you got a point here. You are right in a sense where I shouldn't mixe theming and types styleable.

So we should go like you suggest. And then we will treat the theme in a different topic.

Nexucis:
linked to #5

PR #8, opened 2020-07-17
Implement autocompletion using LSP feature

Nexucis:
linked to #5


src/lang-promql/lsp/model.ts
```diff
@@ -0,0 +1,21 @@
+// TODO consider to use the interface coming from LSP
+// https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion
+export interface TextEdit {
```
slrtbtfs, line 3: This package should provide the types used here: https://www.npmjs.com/package/vscode-languageserver-types

src/lang-promql/lsp/client.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  autocomplete(context: AutocompleteContext): Promise<CompletionResult> {
+    const {state, pos} = context
+    const line = state.doc.lineAt(pos)
+    const body = {
+      expr: state.doc.sliceString(0),
+      limit: 100,
+      // positionChar start at 0 at the beginning of the line for LSP
+      positionChar: pos - line.from - 1 > 0 ? pos - line.from - 1 : pos - line.from,
```
slrtbtfs, line 25: Are you sure these are counted in UTF-16 Codepoints?

If not, the positions will be off in case the query includes non-ASCII characters.

src/lang-promql/lsp/client.ts
```diff
@@ -0,0 +1,61 @@
...
+      // positionLine start at 0 from LSP and at 1 from CMN
+      positionLine: line.number - 1,
+    }
+    return axios.post<AutocompleteResponse[]>(this.url + this.autocompleteEndpoint, body)
+      .then((response) => {
+        const options: Completion[] = []
+        // for every textEdit present, they qll have exactly the same value
```
slrtbtfs, line 32: These TextEdits aren't necessarily all the same.

For example, if a completion result is a function, the completion will include parenthesis and the TextEdit will instruct the client to place the cursor between those. For a metric name no such thing is done.

Nexucis: that doesn't really matter actually. Here we are only considering the position `start` not the position `end`.
And the `start` position is always the same

Maybe I should change a bit my comment, since it's not super clear

src/lang-promql/lsp/client.ts
Nexucis: no idea actually, But if it's not the case, I believe the bug would be more on CodeMirror side

src/lang-promql/lsp/model.ts
Nexucis: I'm not sure yet if we should use it or not. I guess it will depend if we succeed to cover all lsp feature or not.
I don't really want to import a potentially big dependency for just one interface needed

Nexucis: let's see

src/lang-promql/lsp/client.ts
slrtbtfs: Not necessarily.

There is an incredible number of ways to encode positions, and UTF16 Codepoints is one of the weirder ones. Unfortunately that's what the LSP uses.

It could well be (I haven't looked it up), that Codemirror uses a different way of counting chars.

src/lang-promql/lsp/model.ts
slrtbtfs: After passing the Typescript compiler and maybe a Minifier, all the unnecessary Code will probably be optimized away, anyway.

And these types are quite complex. Copying them doesn't seem like a good idea to me.

src/lang-promql/lsp/client.ts
slrtbtfs: I just had a quick look at Codemirror's code, and while it doesn't appear to be documented anywhere, they indeed use UTF16 code points.

src/lang-promql/lsp/client.ts
```diff
@@ -0,0 +1,61 @@
...
+        // and here https://github.com/microsoft/TypeScript/issues/9568
+        let textEdit: TextEdit | undefined
+        if (isIterable(response.data)) {
+          for (const res of response.data) {
+            // TODO map kind with icon Completion.type when it is released on CMN side
+            // https://github.com/codemirror/codemirror.next/commit/459bba29c1fd1d80fc4f36dac27e5825b2362273
+            let label = res.label
+            if (res.textEdit) {
+              label = res.textEdit.newText
+              textEdit = res.textEdit
```
slrtbtfs, line 45: The label and the newText aren't always equal. In this example, they aren't.

```json
[
  {
    "label": "label_replace",
    "kind": 3,
    "sortText": "__1__000016426",
    "insertTextFormat": 2,
    "textEdit": {
      "range": {
        "start": {
          "line": 0,
          "character": 0
        },
        "end": {
          "line": 0,
          "character": 9
        }
      },
      "newText": "label_replace($1)"
    },
    "command": {
      "title": "",
      "command": "editor.action.triggerParameterHints"
    }
  }
]
```

src/lang-promql/lsp/client.ts
Nexucis: nice, thanks !

src/lang-promql/lsp/client.ts
Nexucis: mmm yeah that's the whole point of this code. newText is more accurate than the label. So if exist, then we will always prefer it.
So far I don't see where is the issue.

src/lang-promql/lsp/model.ts
Nexucis: well tree shaking is more on webpack side actually. But yes I see your point here.

Nexucis:
So here is a first version of the autocompletion feature.

I'm providing here a factory in order to be able to have different implementation of the Autocompletion feature.

How it looks like when lsp is activated:

lsp

But plenty of things are still remaining like you can see following the number of TODO (lol):
* Find a nice way to provide the url of the distant server
* Find a nice way to provide the boolean `onlineMode`
* implement a short offline autocomplete
* waiting the release of CMN in order to have the icon in the completion list
* find a better way to handle the server error (today there is none)

Definitely we need to be able to sort properly the list using a scoring system.

@juliusv : maybe you will find some inspiration in this PR specially the way used to contact the distant server. Or maybe not and you already have something better ;).

slrtbtfs:
Nice!

Another To-do:

* The `$1` displayed a few times in the screencast is a placeholder that tells where the cursor should be place and should be used as such.

* If possible: Display the help text in the completion drop down.

src/lang-promql/lsp/client.ts
slrtbtfs: The `label` is what the editor should show in the completion dropdown.

The `newText` describes, what should be inserted, using a template format specified by the LSP.

Showing the `newText` as a `label` violates the LSP.

Nexucis: ok, I managed to do it using the field `apply` from the `Completion`.
Just keep in mind, what we are doing here is just an approximation of the LSP implementation. It can happen we won't do it like LSP is proposing to do ;).

How the display and the feeling with the autocompletion will be, is more important than respecting what LSP is exposing. (at least for me)

Nexucis: Anyway thanks for the explanations, it was interesting to know the different between the two fields.

slrtbtfs: This still doesn't handle the cursor position hints correctly.

It seems to be possible to use a function in the `apply` field which should make it possible to correctly place the cursor after inserting a completion.

slrtbtfs:
Regarding showing documentation while completing, the respective upstream issue seems to be https://github.com/codemirror/codemirror.next/issues/226.

src/lang-promql/lsp/client.ts
slrtbtfs: As long as we're not parsing TextEdits correctly, it's still better to use the `label` field instead of the `newText` for inserting.

That's how Vim handles it, too.

Nexucis: I prefer to use the newText for the moment, because it provides the bracket automatically and so it makes a good difference immediately between a function and a simple string like a label or something equivalent.
Let's see later how to change the cursor position. If we cannot change it, then at least the `$1` will be removed manually.

slrtbtfs: Maybe it's best to leave it as is for now and open an issue to eventually do it properly.

Nexucis: yeah. Here I'm more let's say in a "die and retry" mode. Meaning that I will try many things and then see if the feeling is nice or not.

You can expect that the code is going to change a lot

slrtbtfs:
As far as I'm concerned, we could merge this PR as is and handle the various remaining issues in follow ups.

src/lang-promql/lsp/client.ts
Nexucis: in the meantime, code is opened and the PR as well. Feel free to fix it in case you would like to go furthermore and to enter in the nice world of CodeMirror !

Nexucis:
On my side, the TODOs were more here to raise the different kind of issue I spotted. I didn't have the intention to fix it right now but far later. I should have say it explicitely my bad ^^.

src/lang-promql/lsp/client.ts
slrtbtfs: >  Feel free to fix it in case you would like to go furthermore and to enter in the nice world of CodeMirror !

That'd mean I'd have to actually learn JavaScript ;)

In case I'm be working on this, I think I'll focus more on the Codemirror upstream for now, since two people making large changes to the same code in parallel is likely to get messy.

juliusv:
Regarding placeholders, CM autocomplete has snippet support with tabbable fields, like this:

snippets

For just the bare function names, I don't think we should be autocompleting more than the names. This would get especially annoying if you are updating a function name (like `rate()` -> `irate()`) and the parentheses are already there.

src/lang-promql/complete.ts
```diff
@@ -0,0 +1,117 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import { AutocompleteContext, Completion, CompletionResult } from "@codemirror/next/autocomplete";
+import { LSPBody, LSPClient } from "./lsp/client";
+import { CompletionItem, TextEdit } from "vscode-languageserver-types";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isIterable(obj: any): boolean {
+  return obj != null && typeof obj[ Symbol.iterator ] === 'function';
```
juliusv, line 29: How about using only the strict comparison operators, `!==` instead of `!=`, otherwise `null` also matched `undefined` (and if that's intended, then it's better to make it explicit IMO).

src/lang-promql/complete.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Complete is the interface that defines the simple method that returns a CompletionResult
```
juliusv, line 32: Periods after the end of a sentence (like in Prometheus) will make these comments easier to read.

src/lang-promql/complete.ts
```diff
@@ -0,0 +1,117 @@
...
+// Every different completion mode must implement this interface
+interface Complete {
+  promQL(context: AutocompleteContext): Promise<CompletionResult> | CompletionResult | null;
+}
+
+// LSPComplete will proviode an autocompletion based on what the langserver-promql is providing when requested
```
juliusv, line 38:
```suggestion
// LSPComplete will provide an autocompletion based on what the langserver-promql is providing when requested.
```

src/lang-promql/complete.ts
```diff
@@ -0,0 +1,117 @@
...
+    this.lspClient = lspClient;
+  }
+
+  promQL(context: AutocompleteContext): Promise<CompletionResult> {
+    const {state, pos} = context
+    const line = state.doc.lineAt(pos)
+    const body: LSPBody = {
+      expr: state.doc.sliceString(0),
+      limit: 100,
+      // positionChar start at 0 at the beginning of the line for LSP
```
juliusv, line 52:
```suggestion
      // positionChar starts at 0 at the beginning of the line for LSP.
```

src/lang-promql/complete.ts
```diff
@@ -0,0 +1,117 @@
...
+      positionChar: pos - line.from - 1 > 0 ? pos - line.from - 1 : pos - line.from,
+      // positionLine start at 0 from LSP and at 1 from CMN
```
juliusv, line 54:
```suggestion
      // positionLine starts at 0 from LSP and at 1 from CMN.
```

src/lang-promql/complete.ts
```diff
@@ -0,0 +1,117 @@
...
+      .then((items: CompletionItem[]) => {
+        const options: Completion[] = []
+        // for every `textEdit` present, they all have exactly the same `range.start` value.
+        // so the goal is to save the last one present in order to use it later to calculate the position `from`
+        // Note: textEdit can be undefined but that's just because otherwise it doesn't compile
+        // see more here: https://github.com/microsoft/TypeScript/issues/12916
+        // and here https://github.com/microsoft/TypeScript/issues/9568
+        let textEdit: TextEdit | undefined
+
+        if (isIterable(items)) {
```
juliusv, line 67: In which situations would this not be iterable?

src/lang-promql/complete.ts
```diff
@@ -0,0 +1,117 @@
...
+          from: textEdit ? line.from + textEdit.range.start.character : line.from,
+          to: pos,
+          options: options
+        } as CompletionResult
+      })
+  }
+}
+
+// OfflineComplete is going to provide a full completion result without any distant server
+// So it will basically only provide the different keyword of the PromQL syntax
```
juliusv, line 96: There might also be a use case for an autocompletion of metric names that doesn't go back and forth to the server for every keystroke? Since the metric names can be loaded once upon page load, and then you can do nice and quick fuzzy matching against them, and the grammar tells us when we are completing a metric name.

Nexucis: yes I'm thinking also of an hybrid mode let's say. Since LSP will be available in a future version of Prometheus, then we should have a mode when LSP is not available, to be a bit retro compatible with the different version of prometheus

src/lang-promql/complete.ts
Nexucis: when the server is returning "null" as a response for example.

slrtbtfs:
> For just the bare function names, I don't think we should be autocompleting more than the names. This would get especially annoying if you are updating a function name (like `rate()` -> `irate()`) and the parentheses are already there.

That's something that should be handled by the language server. Currently, when changing the function name, the whole function call is replaced.

src/lang-promql/complete.ts
Nexucis: yes that's intended, I was hesitated by just do it like this:

```typescript
return obj && typeof obj[ Symbol.iterator ] === 'function'
```

but on stackoverflow they all check if it's null instead of doing like the way I wanted. So I let it like that.

src/lang-promql/complete.ts
juliusv: When I see a call like `isIterable()`, I immediately start wondering what other types might be returned, and why... it's fine for now, but if e.g. `null` is the only other valid return type, it could be clearer to just directly mention that, and do `if (items !== null)`, and otherwise assume it's an array.

slrtbtfs:
> Regarding placeholders, CM autocomplete has snippet support with tabbable fields, like this:

So we'd have to translate the LSP snippet format to the CM snippet format?

juliusv:
> So we'd have to translate the LSP snippet format to the CM snippet format?

I guess so! I haven't looked at the LSP one in more detail, but is it just `$<number>`? That doesn't allow nice placeholder names I guess?

src/lang-promql/complete.ts
slrtbtfs: Having a hybrid mode sounds hard to get right, since it's unclear, when the server should be called. The completions that are most expensive to compute are metric names and labels and that's something a offline mode can't do anyway.

What seems more useful to me is caching API responses, as long as the user is typing the same word.

slrtbtfs:
> That doesn't allow nice placeholder names I guess?

It does: https://microsoft.github.io/language-server-protocol/specification.html#snippet_syntax

juliusv:
> It does: https://microsoft.github.io/language-server-protocol/specification.html#snippet_syntax

Nice... yeah, should be pretty straightforward to translate that to CMN's format then. At least the simple format shouldn't even need translation (https://codemirror.net/6/docs/ref/#autocomplete.snippet).

slrtbtfs:
Great!

Regarding the snippet you showed in the screen cast: Do you have a list of useful PromQL snippets? If yes: Would you be willing to open source them (if they aren't already), so they could be added to the language server as well?

src/lang-promql/complete.ts
Nexucis: yeah now your are saying that, maybe it doesn't worth it and we could have something simpler. I think I will just check if the data null or not then

juliusv:
Yeah! I mean, I only experimented with it myself so far, so this is all I've got, feel free to reuse and adapt:

```typescript
export const snippets: readonly SnippetSpec[] = [
  {
    keyword: 'sum(rate(<input vector>[5m]))',
    snippet: 'sum(rate(${<input vector>}[5m]))',
  },
  {
    keyword: 'histogram_quantile(<quantile>, sum by(le) (rate(<histogram metric>[5m])))',
    snippet: 'histogram_quantile(${<quantile>}, sum by(le) (rate(${<histogram metric>}[5m])))',
  },
  {
    keyword: 'label_replace(<input vector>, "<dst>", "<replacement>", "<src>", "<regex>")',
    snippet: 'label_replace(${<input vector>}, "${<dst>}", "${<replacement>}", "${<src>}", "${<regex>}")',
  },
];
```

slrtbtfs:
Cool!

I'll see if something like this can be added to the language server.

src/lang-promql/complete.ts
juliusv: Yeah, I'm also saying that because we also have an [`isPresent()` function](https://github.com/prometheus/prometheus/blob/490f9c664e48fbd1509c42ae85fa566901b11a99/web/ui/react-app/src/utils/index.ts#L14) in Prometheus, and any time I see it used, I stumble and ask myself: what are the actually possible / expected values here, and what *exactly* does this function do? So maybe better to leave it less abstracted in the beginning :)

src/lang-promql/complete.ts
Nexucis: caching is just an optimization. It will be handled by a `prometheusClient` or by the hybidMode itself. 

For example like that: 

```typescript
class HybridComplete implements Complete {
  private metrics: string[];
  promQL(context: AutocompleteContext): Promise<CompletionResult> {
      const {state, pos} = context
     // Analyze the tree.
     // When required call the prometheusClient
     if (metrics.lenght) <= 0 {
         this.promClient.getMetricName()
           .then((result) => {
             this.metrics = result
             // tranform the result to the correct format
           })
     }
  }
}

**Comment by juliusv** [comment]
Ah interesting, adding that straight to the language server.

Note that CMN will only handle snippets automatically when they are provided via `completeSnippets()`, otherwise I guess you have to detect use `snippet()` (https://codemirror.net/6/docs/ref/#autocomplete.snippet) to parse the snippet and get a function that applies it to the editor state.

File: src/lang-promql/complete.ts (hunk 11 above)
Comment (Nexucis) [reply src/lang-promql/complete.ts:96]: I think this mode should able to replace properly what you are doing with Monaco @juliusv

Comment (juliusv) [reply src/lang-promql/complete.ts:96]: @Nexucis To fully replace what I get "out of the box" with Monaco, I would need:

* Fast offline fuzzy matching against once-loaded metric names (good, with proper scoring, etc.) (https://github.com/codemirror/codemirror.next/issues/236)
* Highlighting of partial fuzzy match results (https://github.com/codemirror/codemirror.next/issues/234)
* Icons that indicate the completion item type (https://github.com/codemirror/codemirror.next/issues/226)

Comment (slrtbtfs) [reply src/lang-promql/complete.ts:96]: > caching is just an optimization. It will be handled by a prometheusClient or by the hybidMode itself.

I was talking about caching the langserver API responses and then applying offline fuzzy matching. That is nothing that is specific to metric name completions.

In fact, when using the language server API, CM doesn't need to be concerned with what kind of completion to look for.

Comment (slrtbtfs) [reply src/lang-promql/complete.ts:96]: I'm still not sure what problems a hybrid mode, that does more than caching and fuzzy matching, would solve.

Comment (Nexucis) [reply src/lang-promql/complete.ts:96]: Yeah I know don't worry. I was more talking about what you have with the auto-completion in term of data. Rendering and visualization is another part.
* For the icons, it's just matter of time since it's already implemented, we just have to wait the next release of codemirror.
* For the fuzzy matching, well as you can see I'm waiting the answer of marijnh. So I didn't have better idea, and it should be enough to cover the needs IMHO.
* For the hightlighting, I'm still searching how to do it. I think that's the part that will take me the most time. Others should be quite easier, just matter of algorithm and looping on struct ^^

Comment (Nexucis) [reply src/lang-promql/complete.ts:96]: > I'm still not sure what problems a hybrid mode, that does more than caching and fuzzy matching, would solve.

It will be a short replacement of LSP when it's not available. I'm saying a "short" replacement, because for sure it won't be possible/easier to implement all functionnality of the completion provided by LSP.

File: src/lang-promql/complete.ts (hunk 10 above)
Comment (Nexucis) [reply src/lang-promql/complete.ts:67]: So far instead of calling `isPresent`, I would just replace it by 

```typescript
if (myObject) {
}
```

Should normally work in every situation; but that's just my opinion :). That's what I'm going to do here

juliusv: I hate the concept of "falsy" in languages, but ok :)

Nexucis: hahaha. For me that just a specificity of the language like every language has. But I can understand this repulsion against this thing :).

Nexucis:
Thanks for the review from both of you @juliusv and @slrtbtfs. 
There will be for sure other PR to improve the LSP mode following all what you said :). 
Just keep in mind, that's just the beginning. On my side I started by this part, because that was a good start to make deep dive of how the completion is working in CMN.

Cheers!

juliusv:
Yeah, I think it's ok to merge for now, and then we can iteratively improve.

PR #11, opened 2020-08-01
Improve Hybrid mode + add icon support for LSP

Nexucis:
## Icon supported
Last version of the CMN (v0.9.0) is supporting the icon for the autocompletion.
Icon is now included for LSP and the Hybrid mode.

## Cache system
A cache system is now included for the Hybrid mode.

## Filter all result
the autocompletion list is now filtered according to what the user is taping.

## Associated the label value to the couple <MetricName, LabelName>
previously when you would like to autocomplete a label value, it was done without considering if the given labelName was associated to a metric. Then it was returning some value that wasn't accurate : 

**Before making the association**
before_associated_metric_label_value_name

**After**

after_associated_metric_label_value_name

## Final visualization

cache


Nexucis:
linked to #5

README.md
```diff
@@ -33,16 +33,7 @@ Here is a picture that displays what happen when you write a promQL expression:
...
-
-- [x] Developed an offline mode using a light promQL grammar
-- [ ] Build a better development environment
-  - [x] Create a test folder in the src folder that will contain a light app that will start the codeMirror
-  - [ ] Automatized test to checked that all correct promQL syntax are covered
-- [ ] Implement the Online mode
-- [ ] Offline mode improvement
-  - [ ] Improve the error returns when a wrong promQL syntax is entered
-  - [ ] Use directly the parser generated by the grammar to be able to add the check done in the [promQL parser](https://github.com/prometheus/prometheus/blob/fac7a4a0504404fa5d4c5abb8fcc9750bd5cbda7/promql/parser/parse.go#L510-L515)
+Roadmap is available in the issue [#5](https://github.com/prometheus-community/codemirror-promql/issues/5)
```
juliusv, line 36: I think you can just say `in the issue #5` and it will auto-link :)

juliusv:
I think rather than me reviewing everything in detail here without trying it out, let's merge it for now and iterate.

juliusv:
(and looks awesome from the GIFs already!)

juliusv:
(hit the wrong button)

PR #14, opened 2020-08-04
Switch to Prometheus ESLint config

juliusv:
This switches the ESLint config to the one that Prometheus uses, minus the React bits. Not every lint+prettier rule will be everyone's favorite and might take some getting-used-to, but it'll be good to have a consistent style with Prometheus :)


Nexucis:
So far, it looks like the same coding style than the one you can find in Angular. Sounds good to me, I like these rules :D

Nexucis:
thanks for the PR !

PR #16, opened 2020-08-06
add material based theme

Nexucis:
image

image

/cc @juliusv

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -21,13 +21,12 @@
 // SOFTWARE.
 
 import { EditorState, EditorView } from '@codemirror/next/basic-setup';
-import { promQL, setComplete } from '../lang-promql';
+import { promQL, promQLHighlightMaterialTheme, setComplete } from '../lang-promql';
 import { Extension } from '@codemirror/next/state';
 import { history, historyKeymap } from '@codemirror/next/history';
 import { highlightSpecialChars, keymap, multipleSelections } from '@codemirror/next/view';
 import { lineNumbers } from '@codemirror/next/gutter';
 import { foldGutter, foldKeymap } from '@codemirror/next/fold';
-import { defaultHighlighter } from '@codemirror/next/highlight';
 import { bracketMatching } from '@codemirror/next/matchbrackets';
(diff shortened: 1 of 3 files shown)


Nexucis:
well it looks like Circle-ci has some issues with npm. I will have to re-run the build a bit later: 
image

slrtbtfs:
Looking at the screenshot, the green on white looks barely readable to me. Could this be replaced by a darker tone?

Nexucis:
better ? 
image

slrtbtfs:
Yes.

Nexucis:
related to #5

juliusv:
Btw., this was my theme to make it look like Monaco / promlens.com:

```typescript
{
      name: { color: '#000' },
      number: { color: '#09885a' },
      string: { color: '#a31515' },
      keyword: { color: '#008080' },
      labelName: { color: '#800000' },
      operator: {},
      modifier: { color: '#008080' },
      paren: {},
      squareBracket: {},
      brace: {},
      invalid: { color: 'red' },
      comment: { color: '#888', fontStyle: 'italic' },
    }
```

...though that was a while ago, could be that would need an update after the latest CM-promql changes.

Nexucis:
Well it won't hurt to add another one :). Like `promQLHighlightMonacoTheme`.

I'm not really certain theme should be delivered by this lib actually. But well for the moment it won't be a big issue to add it here

juliusv:
I just commented since it might be interesting. Yeah, I don't think themes should go into this repo, or that they should even be tied to a language at all.

Nexucis:
yeah I'm agree, theme should be delivered by CMN directly. Like it is for the previous version.

Since it's not yet available and since I would like to have some nice color for the playground I added this theme. So that mean, if you want we can add as well yours :)

juliusv:
If you want a nice theme for the playground, I would propose just putting that theme code directly into the playground example code, rather than into the `lang-promql` subdir?

Nexucis:
mmm yeah makes sense, I will change that. Thanks for the suggestion :)

Nexucis:
regarding the theme you posted, yeah it requires some adjustment if we want to use it in Prometheus/ PromLabs, but nothing too serious.

PR #18, opened 2020-08-07
highlight string matching during autocompletion

Nexucis:
final_highlight

Should complete the minimum you wanted @juliusv :D

related to #5 

Note: Behind the seen, it is used what I proposed to CMN: https://github.com/codemirror/codemirror.next/issues/234

### DIFF
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -76,15 +76,15 @@ const autocompleteNode = {
 
 const snippets: readonly SnippetSpec[] = [
   {
-    keyword: 'sum(rate(&lt;input vector&gt;[5m]))',
+    keyword: 'sum(rate(<input vector>;[5m]))',
     snippet: 'sum(rate(${<input vector>}[5m]))',
   },
   {
-    keyword: 'histogram_quantile(&lt;quantile&gt;, sum by(le) (rate(&lt;histogram metric&gt;[5m])))',
+    keyword: 'histogram_quantile(<quantile>, sum by(le) (rate(<histogram metric>;[5m])))',
     snippet: 'histogram_quantile(${<quantile>}, sum by(le) (rate(${<histogram metric>}[5m])))',
   },
   {
-    keyword: 'label_replace(&lt;input vector&gt;, ";&lt;dst&gt;", "&lt;replacement&gt;", "&lt;src&gt;", "&lt;regex&gt;")',
+    keyword: 'label_replace(<input vector>;, "<dst>", "<replacement>", "<src>", "<regex>")',
     snippet: 'label_replace(${<input vector>}, "${<dst>}", "${<replacement>}", "${<src>}", "${<regex>}")',
   },
 ];
@@ -96,16 +96,20 @@ const parsedSnippets = snippets.map((s) => ({
   score: 0,
 }));
 
+function escape(text: string): string {
+  return text.replace(/</g, '&lt;').replace(/>/g, '&gt;');
+}
+
 // HybridComplete is going to provide a full completion result with or without a remote prometheus.
 export class HybridComplete implements CompleteStrategy {
   private readonly prometheusClient: PrometheusClient | null;
-  private fuzzyPre: string;
-  private fuzzyPost: string;
+  private readonly fuzzyPre: string;
+  private readonly fuzzyPost: string;
 
-  constructor(prometheusClient: PrometheusClient | null, fuzzyPre: string | undefined, fuzzyPost: string | undefined) {
+  constructor(prometheusClient: PrometheusClient | null, fuzzyPre?: string, fuzzyPost?: string) {
     this.prometheusClient = prometheusClient;
-    this.fuzzyPre = fuzzyPre ? fuzzyPre : '';
-    this.fuzzyPost = fuzzyPost ? fuzzyPost : '';
+    this.fuzzyPre = fuzzyPre || '';
+    this.fuzzyPost = fuzzyPost || '';
   }
 
   promQL(context: AutocompleteContext): Promise<CompletionResult> | CompletionResult | null {
@@ -269,14 +273,14 @@ export class HybridComplete implements CompleteStrategy {
     for (const completionList of data) {
       for (const label of completionList.labels) {
         if (context.filterType === FilterType.Fuzzy) {
-          const result = fuzzy.filter(text, [label], fuzzyOption);
+          const result = fuzzy.filter(text, [escape(label)], fuzzyOption);
           if (result && result.length > 0) {
             options.push({ label: result[0].string, apply: label, type: completionList.type, score: result[0].score });
           }
         } else {
           let score: number | null;
           if ((score = context.filter(label, text, true))) {
-            options.push({ label: label, apply: label, type: completionList.type, score: score });
+            options.push({ label: escape(label), apply: label, type: completionList.type, score: score });
           }
         }
       }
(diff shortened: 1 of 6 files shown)


src/lang-promql/complete/hybrid.ts
```diff
@@ -281,4 +253,54 @@ export class HybridComplete implements CompleteStrategy {
...
+    includeSnippet = false
+  ): CompletionResult {
+    const text = state.sliceDoc(from, to);
+    const options: Completion[] = [];
+    const fuzzyOption = { pre: this.fuzzyPre, post: this.fuzzyPost };
+
+    for (const completionList of data) {
+      for (const label of completionList.labels) {
+        if (context.filterType === FilterType.Fuzzy) {
+          const result = fuzzy.filter(text, [label], fuzzyOption);
```
juliusv, line 272: Somewhere, we will have to HTML-escape label names (since they can contain arbitrary strings).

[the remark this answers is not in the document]
Nexucis: I have no idea. At the beginning, I was passing the complete list, but it was simpler to passing them one by one actually.

But if it's the case, it won't hurt to change it :)

Nexucis: should be fine now ;)

src/lang-promql/complete/hybrid.ts
```diff
@@ -94,46 +96,16 @@ const parsedSnippets = snippets.map((s) => ({
...
 // HybridComplete is going to provide a full completion result with or without a remote prometheus.
 export class HybridComplete implements CompleteStrategy {
   private readonly prometheusClient: PrometheusClient | null;
+  private fuzzyPre: string;
+  private fuzzyPost: string;
 
-  constructor(prometheusClient: PrometheusClient | null) {
+  constructor(prometheusClient: PrometheusClient | null, fuzzyPre: string | undefined, fuzzyPost: string | undefined) {
     this.prometheusClient = prometheusClient;
+    this.fuzzyPre = fuzzyPre ? fuzzyPre : '';
```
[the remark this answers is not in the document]
Nexucis: fancy Typescript syntax hahaha. Thanks for the tips

src/lang-promql/complete/hybrid.ts
```diff
@@ -94,46 +96,16 @@ const parsedSnippets = snippets.map((s) => ({
...
-}
-
 // HybridComplete is going to provide a full completion result with or without a remote prometheus.
 export class HybridComplete implements CompleteStrategy {
   private readonly prometheusClient: PrometheusClient | null;
+  private fuzzyPre: string;
+  private fuzzyPost: string;
 
-  constructor(prometheusClient: PrometheusClient | null) {
+  constructor(prometheusClient: PrometheusClient | null, fuzzyPre: string | undefined, fuzzyPost: string | undefined) {
```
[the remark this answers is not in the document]
Nexucis: ah yeah that's definitely simpler. Thanks!

src/lang-promql/complete/hybrid.ts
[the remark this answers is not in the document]
juliusv: That's just normal JS syntax :)

src/lang-promql/complete/hybrid.ts
```diff
@@ -94,46 +96,20 @@ const parsedSnippets = snippets.map((s) => ({
...
-      }
-    }
-  }
-  return {
-    from: from,
-    to: to,
-    options: options,
-  } as CompletionResult;
+function escape(text: string): string {
+  return text.replace(/</g, '&lt;').replace(/>/g, '&gt;');
```
juliusv, line 100: You'll need to replace at least `&` as well, and maybe better also a couple of other chars. This is the `escapeHTML()` function from Prometheus:

https://github.com/prometheus/prometheus/blob/f97d2ddb6e6fce1343d2b026479d19592b934a36/web/ui/react-app/src/utils/index.ts#L16-L29

Nexucis: Ah ok. Funny things btw, looks like the map is worst choice for this use case.
https://stackoverflow.com/a/28458409

No Idea how the Switch case ban be better than a map.
I supposed you lost time when creating the map

src/lang-promql/complete/hybrid.ts
```diff
@@ -281,4 +257,53 @@ export class HybridComplete implements CompleteStrategy {
...
+    includeSnippet = false
+  ): CompletionResult {
+    const text = state.sliceDoc(from, to);
+    const options: Completion[] = [];
+    const fuzzyOption = { pre: this.fuzzyPre, post: this.fuzzyPost };
+
+    for (const completionList of data) {
+      for (const label of completionList.labels) {
+        if (context.filterType === FilterType.Fuzzy) {
+          const result = fuzzy.filter(text, [escape(label)], fuzzyOption);
```
juliusv, line 276: Btw., if you're passing just one element at a time, we are not using half of `fuzzy.filter()`'s functionality, because it also does sorting by score. You can probably just copy this bit right here:

https://github.com/mattyork/fuzzy/blob/master/lib/fuzzy.js#L120-L129

...with the thing in the `if` block just being a different variant of what you already have here.

src/lang-promql/complete/hybrid.ts
juliusv: Interesting... though the differences are so minor that it probably doesn't matter. The worst case (several regex replaces) is "2% slower" in my browser.

src/lang-promql/complete/hybrid.ts
Nexucis: sorting is perfomed by CMN after (using the score). So it's fine to do that from my point of view

Nexucis: I'm just using fuzzy lib here to calculate the score and to inject the html when configured

juliusv: Yeah, exactly. Since we are doing sorting externally, we could just extract the tiny bit from `fuzzy.filter()` that we actually need here.

juliusv: I.e. just use `fuzzy.match()` directly instead of `fuzzy.filter()`.

Nexucis: aaah ok right. Sorry, it was bit confused in my head.

src/lang-promql/complete/hybrid.ts
Nexucis: yup it doesn't really matter, I just found it funny ;)

src/lang-promql/complete/hybrid.ts
Nexucis: oh yeah it's so much better. Thanks a lot for the tips !

Nexucis: here we go, fixed !

src/lang-promql/complete/hybrid.ts
Nexucis: fixed

src/lang-promql/complete/hybrid.ts
```diff
@@ -74,15 +76,15 @@ const autocompleteNode = {
 
 const snippets: readonly SnippetSpec[] = [
   {
-    keyword: 'sum(rate(<input vector>[5m]))',
+    keyword: 'sum(rate(&lt;input vector>[5m]))',
```
juliusv, line 79: I think you I preferred how you had these snippets non-escaped a few commits ago. If we always just escape it during the autocomplete, then when you define snippets here, you don't have to worry about HTML escaping (or can do it wrong and forget it).

Nexucis: It's actually in order to avoid to match the string that replaces the double quote.
Because well if you are typing `sum` for example, then if everything is escaped, then the snippet `label_replace` came up.

Nexucis: Well I think, the only way to have an accurate match will have to escape after the fuzzy. I don't have other solution to be sure that it won't match the escape characters :/

juliusv: Ok yeah, then let's leave it like this for now.

It's a bit of a pity that the fuzzy lib is unmaintained since 4 years, otherwise I would propose them to add an option to allow escaping for `ch` here: https://github.com/mattyork/fuzzy/blob/master/lib/fuzzy.js#L60

Maybe time for another fork sometime ;)

juliusv:
:+1: Thanks! :)

PR #19, opened 2020-08-08
Make the completion extension state not global

piotaixr:
Signed-off-by: Rémi Piotaix <remi@piotaix.fr>


Nexucis:
awesome, thanks ! That helps a lot :)

PR #21, opened 2020-08-09
Integrate HTTP Error Handler + factorize the way to filter the result

Nexucis:
related to #5


Nexucis:
should fix the issue you had @juliusv when prometheus wasn't reachable.

juliusv:
:+1: Thanks! :)

PR #23, opened 2020-08-11
Adapt code to autocomplete module changes

juliusv:
This is to adjust the usage for the changes in https://github.com/Nexucis/codemirror-next-autocomplete/pull/1

### DIFF
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -24,16 +24,14 @@ import {
   AutocompleteContext,
   Completion,
   CompletionResult,
-  FilterType,
   snippet,
   SnippetSpec,
 } from '@nexucis/codemirror-next-autocomplete';
 import { CompleteStrategy } from './index';
 import { PrometheusClient } from './prometheus/client';
 import { Subtree } from 'lezer-tree';
-import { EditorState, EditorView } from '@codemirror/next/basic-setup';
+import { EditorState } from '@codemirror/next/basic-setup';
 import { promQLSyntax } from 'lezer-promql';
-import * as fuzzy from 'fuzzy';
 
 interface AutoCompleteNode {
   labels: string[];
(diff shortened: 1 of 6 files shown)


juliusv:
Note that this depends on the linked PR in the autocomplete module and will need a package.json dependency update for that before working here. But then it should look like this:

noflicker_autocomplete

src/lang-promql/complete/hybrid.ts
```diff
@@ -284,16 +262,20 @@ export class HybridComplete implements CompleteStrategy {
 
     for (const completionList of data) {
       for (const label of completionList.labels) {
-        const completionResult = this.filter(escape(label), text, label, context);
-        if (completionResult) {
+        const completionResult = context.filter(
+          { label: label, original: label, apply: label, type: completionList.type, score: 0 },
+          text
+        );
+        console.log('==', completionResult);
```
Nexucis, line 269: I guess this should be emoved. Quite weird eslint doesn't yell at it

src/lang-promql/complete/hybrid.ts
```diff
@@ -302,29 +284,7 @@ export class HybridComplete implements CompleteStrategy {
       from: from,
       to: to,
       options: options,
+      filterDownOn: /^[a-zA-Z0-9_:]+$/,
```
Nexucis, line 287: Maybe we should make it optional ?

src/lang-promql/complete/hybrid.ts
```diff
@@ -284,16 +262,20 @@ export class HybridComplete implements CompleteStrategy {
...
           options.push(completionResult);
         }
       }
     }
     if (includeSnippet) {
       for (const s of parsedSnippets) {
-        const completionResult = this.filter(s.label, text, s.apply, context);
-        if (completionResult) {
+        const completionResult = context.filter(s, text);
+        if (completionResult !== null) {
```
Nexucis, line 278: You really don't like the simple check :
```typescript
if (completionResult) {
}
```

src/lang-promql/complete/hybrid.ts
juliusv: Whoops, thanks, fixed! :)

src/lang-promql/complete/hybrid.ts
juliusv: Hm we could, maybe in a followup version? We also have to rethink the characters maybe, since they would depend on the type of object being completed...

src/lang-promql/complete/hybrid.ts
juliusv: Yeah, that's just the wild west to me, that could mean anything ;)

Nexucis: Guess it's a question of habits. But ok it's fine

src/lang-promql/complete/hybrid.ts
Nexucis: Yeah ok. It will give us time to see if it's a good idea or something that should stay in the garbage 1 am Idea hahaha

PR #24, opened 2020-08-13
Fix autocomplete parameters in demo app

juliusv:
Also fixes the main package's version in package-lock.json.


Nexucis:
Lol I did the same fix at the same time in #10. 
Thanks for the fix 👍 

I should put a CI check to verify that the app is working

PR #26, opened 2020-08-13
upgrade codemirror-next-autocomplete dependency that contains the new fuzzy lib

### LINKED ISSUE #20: Replace lib fuzzy by fuse.js
* [fuzzy](https://github.com/mattyork/fuzzy) looks like it's abandoned and doesn't support ES5
* [fuse.js](https://github.com/krisk/fuse) can return the indices of matched characters. Useful for enhance the highlight and the escape of the HTML characters.

Nexucis:
fix #20

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "codemirror-promql",
-  "version": "0.4.1",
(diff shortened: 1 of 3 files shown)


juliusv:
:+1:

PR #27, opened 2020-08-13
expose PrometheusClient and LSPClient + deep review of the CompleteConfiguration

Nexucis:


src/lang-promql/complete/index.ts
```diff
@@ -34,21 +33,31 @@ export interface CompleteStrategy {
...
-  offline: boolean;
-  url: string;
-  httpErrorHandler?: (error: any) => void;
+  lsp?: {
+    url: string;
+    lspClient?: LSPClient;
+    httpErrorHandler?: (error: any) => void;
+  };
   hybrid?: {
+    url: string;
```
juliusv, line 42: It seems that `url` should be optional now, or rather that exactly one of `url` or `prometheusClient` should be provided (and that an error is raised otherwise).

To be honest to keep the interface cleaner, I would probably remove the option of passing a URL and always require a Prometheus client to be passed in. But it's a bit of a matter of taste, and of course that means that each caller has a tiny bit more work to do.

Nexucis: yeah you are right. URL should be optional.

For the 2nd point, I prefer to have an easy conf to setup. From my point of view, adding a new instance of the `PrometheusClient` is a bit advance. It wouldn't be nice for the basic use case you have to instantiate the default `PrometheusClient` by yourself

juliusv: Ok!

Nexucis: done

juliusv:
:+1:

PR #29, opened 2020-08-15
Implement the linter strategy starting by LSP

Nexucis:
lsp-linter

/cc @slrtbtfs 

Offline linter will come later

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -65,47 +65,76 @@ const basicSetup: Extension = [
 ];
 
 const promqlExtension = new PromQLExtension();
+let editor: EditorView;
 
-function activateLSPAutocompletion(): void {
-  promqlExtension.setComplete({
-    lsp: {
-      url: 'http://localhost:8080',
-    },
+function setCompletion() {
+  const completionSelect = document.getElementById('completion') as HTMLSelectElement;
+  const completionValue = completionSelect.options[completionSelect.selectedIndex].value;
+  switch (completionValue) {
+    case 'offline':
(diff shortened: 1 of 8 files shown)


Nexucis:
Related to #5

src/app/app.ts
```diff
@@ -66,46 +66,74 @@ const basicSetup: Extension = [
...
-  activateLSPAutocompletion();
-});
+function applyConfiguration(): void {
+  setCompletion();
+  setLinter();
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
+  // When the linter is changed, it required to reload completely the editor.
+  // So the first thing to do, is to completely delete the previous editor and to recreate it from scratch
+  document.getElementById('editor').innerHTML = '';
```
juliusv, line 129: Might be nicer to save the `EditorView` somewhere and call `view.destroy()` on it here vs. just setting the `innerHTML` to empty?

juliusv:
:+1: besides comment

src/app/app.ts
Nexucis: yup definitely better. Didn't know it exists ( and also I didn't search it my bad )

Thanks for the info :)

Nexucis:
> It might be worth investigating whether it's possible to additionally show those error messages below the editor, without having to hover over the red lines.

No idea if it's possible, but otherwise it's how the IDE are working in generally, and I didn't think it was so bad. But maybe it's just me hahaha

PR #33, opened 2020-08-20
Better autocomplete for modifiers and matchOps / binOps

juliusv:
This isn't perfect and completes too much in some situations, but at least
there is now completion happening at all for match ops, binops, and binop
modifiers in places where they make sense.

### DIFF
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -42,7 +42,8 @@ import {
   MetricIdentifier,
   StringLiteral,
   VectorSelector,
-  AggregateModifier,
+  binOpModifierTerms,
+  aggregateOpModifierTerms,
 } from 'lezer-promql';
 
 interface AutoCompleteNode {
@@ -53,10 +54,10 @@ interface AutoCompleteNode {
 const autocompleteNode: { [key: string]: AutoCompleteNode } = {
   matchOp: { labels: matchOpTerms, type: '' },
   binOp: { labels: binOpTerms, type: '' },
-  binOpModifier: { labels: ['on', 'ignoring', 'group_left', 'group_right'], type: 'keyword' },
+  binOpModifier: { labels: binOpModifierTerms, type: 'keyword' },
   functionIdentifier: { labels: functionIdentifierTerms, type: 'function' },
   aggregateOp: { labels: aggregateOpTerms, type: 'keyword' },
-  aggregateOpModifier: { labels: ['by', 'without'], type: 'keyword' },
+  aggregateOpModifier: { labels: aggregateOpModifierTerms, type: 'keyword' },
 };
 
 const snippets: readonly SnippetSpec[] = [
@@ -98,7 +99,7 @@ export class HybridComplete implements CompleteStrategy {
 
       if (tree.parent?.parent?.parent?.parent?.type.id === BinaryExpr) {
         // This is for autocompleting binary operator modifiers (on / ignoring / group_x). When we get here, we have something like:
-        //       metric_name / igno
+        //       metric_name / ignor
         // And the tree components above the half-finished set operator will look like:
         //
         // Identifier -> MetricIdentifier -> VectorSelector -> Expr -> BinaryExpr.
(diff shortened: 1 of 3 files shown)


src/lang-promql/complete/hybrid.ts
```diff
@@ -42,20 +42,21 @@ import {
...
 
-const autocompleteNode: { [key: number]: AutoCompleteNode } = {
-  [matchOp]: { labels: matchOpTerms, type: '' },
-  [BinaryExpr]: { labels: binOpTerms, type: '' },
-  [FunctionIdentifier]: { labels: functionIdentifierTerms, type: 'function' },
-  [AggregateOp]: { labels: aggregateOpTerms, type: 'keyword' },
+const autocompleteNode: { [key: string]: AutoCompleteNode } = {
+  matchOp: { labels: matchOpTerms, type: '' },
+  binOp: { labels: binOpTerms, type: '' },
+  binOpModifier: { labels: ['on', 'ignoring', 'group_left', 'group_right'], type: 'keyword' },
```
Nexucis, line 56: you don't want to add this list to lezer-promql ? It would be consistent like that

src/lang-promql/complete/hybrid.ts
```diff
@@ -42,20 +42,21 @@ import {
...
+  functionIdentifier: { labels: functionIdentifierTerms, type: 'function' },
+  aggregateOp: { labels: aggregateOpTerms, type: 'keyword' },
+  aggregateOpModifier: { labels: ['by', 'without'], type: 'keyword' },
```
Nexucis, line 59: same comment here

src/lang-promql/complete/hybrid.ts
```diff
@@ -134,17 +132,33 @@ export class HybridComplete implements CompleteStrategy {
       // So we can autocomplete the labelValue
       return this.autocompleteLabelValue(tree.parent, tree, pos, context, state);
     }
+    if (tree.type.id === LabelMatcher && tree.firstChild?.type.id === LabelName && tree.lastChild !== null) {
```
Nexucis, line 135: isn't better like that:
```typescript
tree.lastChild?.name === '/!\' 
```

src/lang-promql/complete/hybrid.ts
```diff
@@ -134,17 +132,33 @@ export class HybridComplete implements CompleteStrategy {
...
+    }
+    if ((tree.type.id === Identifier && tree.parent?.type.id === 0) || (tree.type.id === 0 && tree.parent?.type.id !== LabelMatchers)) {
+      // This matches identifier-ish keywords in certain places where a normal identifier would be invalid, like completing "b" into "by":
+      //        sum b
+      // ...or:
+      //        sum(metric_name) b
+      // ...or completing "unle" into "unless":
+      //        metric_name / unle
+      // TODO: This is imprecise and autocompletes in too many situations. Make this better.
+      return this.arrayToCompletionResult([autocompleteNode.aggregateOpModifier].concat(autocompleteNode.binOp), tree.start, pos, context, state);
```
Nexucis, line 161: here does it make sense to autocomplete the metrics name too ?

src/lang-promql/complete/hybrid.ts
```diff
@@ -93,29 +94,26 @@ export class HybridComplete implements CompleteStrategy {
     const { state, pos } = context;
     const tree = state.tree.resolve(pos, -1);
     if (tree.parent?.type.id === MetricIdentifier && tree.type.id === Identifier) {
+      let nonMetricCompletions = [autocompleteNode.functionIdentifier, autocompleteNode.aggregateOp, autocompleteNode.binOp];
+
+      if (tree.parent?.parent?.parent?.parent?.type.id === BinaryExpr) {
```
Nexucis, line 99: yeah we really need a sort of Tree Walker. Looks like a bit too random and too dependent of lezer-promql.

Nexucis (commented):
Thanks for taking the time to improve it :). It definitely requires a `TreeWalker` at some point to be a bit more robust.

src/lang-promql/complete/hybrid.ts
juliusv: No, these are places where a normal expression (of which a metric name or function call are examples) would be invalid. E.g. you can't have `metric_name1 metric_name2`, that's not a valid PromQL expression.

src/lang-promql/complete/hybrid.ts
juliusv: I changed it to `tree.lastChild?.type.id === 0` instead, like further down below, so I don't have to compare to some weird error emoji :)

I also added another condition to stop autocompleting multiple match ops after another...

src/lang-promql/complete/hybrid.ts
juliusv: Hm ok yeah, I added these to lezer and updated lezer here. I'm still really not happy about those things living there, but I don't know a better solution either for the moment.

src/lang-promql/complete/hybrid.ts
juliusv: Same

src/lang-promql/complete/hybrid.ts
juliusv: Yeah. It will always be dependent on lezer-promql though, even if we can abstract it a bit at some point. Basically we have to guess (or rather, experiment) what Lezer guesses what an expression is supposed to be when it's not a fully valid expression.

juliusv (commented):
Pushed some changes.

Nexucis (approved):
LGTM :)

PR #37, opened 2020-08-27
implement vectorSelector linter

Nexucis:
vectorSelector_linter

### DIFF
--- a/src/lang-promql/parser/matcher.ts
+++ b/src/lang-promql/parser/matcher.ts
@@ -1,0 +1,87 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2020 The Prometheus Authors
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import { Subtree } from 'lezer-tree';
+import { EqlRegex, EqlSingle, LabelMatcher, LabelName, MatchOp, Neq, NeqRegex, StringLiteral } from 'lezer-promql';
+import { EditorState } from '@codemirror/next/basic-setup';
+
+class Matcher {
+  type: number;
+  name: string;
+  value: string;
+
+  constructor(type: number, name: string, value: string) {
+    this.type = type;
+    this.name = name;
+    this.value = value;
+  }
+
+  matchesEmpty(): boolean {
+    switch (this.type) {
+      case EqlSingle:
+      case EqlRegex:
+        return this.value === '';
+      case Neq:
+      case NeqRegex:
+        return this.value !== '';
+      default:
+        return false;
+    }
+  }
+}
+
+function createMatcher(labelMatcher: Subtree, state: EditorState): Matcher {
+  const matcher = new Matcher(0, '', '');
+  labelMatcher.iterate({
+    enter: (type, start, end) => {
+      switch (type.id) {
+        case LabelMatcher:
+          return undefined;
+        case LabelName:
+          matcher.name = state.sliceDoc(start, end);
+          break;
+        case MatchOp:
+          return undefined;
+        case EqlSingle:
+        case Neq:
+        case EqlRegex:
+        case NeqRegex:
+          matcher.type = type.id;
+          break;
+        case StringLiteral:
+          const value = state.sliceDoc(start, end);
+          matcher.value = value === "''" || value === '""' || value === '``' ? '' : value;
+          break;
+      }
+      return false;
+    },
+  });
+  return matcher;
+}
+
+export function buildLabelMatchers(labelMatchers: Subtree[], state: EditorState): Matcher[] {
+  const matchers: Matcher[] = [];
+  labelMatchers.forEach((value) => {
+    matchers.push(createMatcher(value, state));
+  });
+  return matchers;
+}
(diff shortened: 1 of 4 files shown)


juliusv:
`foo{bar=""}` is actually legal, as the `foo` itself is already a non-empty matcher. It's equivalent to `{__name__="foo",bar=""}`

Nexucis:
ah ok. mmm I'm missing something in this algo https://github.com/prometheus/prometheus/blob/master/promql/parser/parse.go#L589 then.

src/lang-promql/parser/matcher.ts
```diff
@@ -0,0 +1,87 @@
...
+  name: string;
+  value: string;
+
+  constructor(type: number, name: string, value: string) {
+    this.type = type;
+    this.name = name;
+    this.value = value;
+  }
+
+  matchesEmpty(): boolean {
```
juliusv, line 38: You won't be able to detect matchers that match the empty string perfectly in Javascript, as you would need a JS-based RE2 (https://github.com/google/re2) implementation.

For example, `foo=~".*"` would also match empty labels although the regex string is not empty, and on the flip side, `foo!~"|bar"` does *not* match empty labels, although the regex string is not empty.

You could try matching the regexes in JS against an empty string (like we do in Go), but since JS regexes are different from RE2 regexes, this won't work perfectly either.

So maybe the best you can do is to only detect some cases, like `EqlSingle` and `Neq` ones, but not the regex cases (you could include some special-case hardcoded regex cases too, but it should never return a linter error for something that is not actually a matcher that matches an empty label).

src/lang-promql/parser/matcher.ts
```diff
@@ -0,0 +1,87 @@
...
+          return undefined;
+        case EqlSingle:
+        case Neq:
+        case EqlRegex:
+        case NeqRegex:
+          matcher.type = type.id;
+          break;
+        case StringLiteral:
+          const value = state.sliceDoc(start, end);
+          matcher.value = value === "''" || value === '""' || value === '``' ? '' : value;
```
juliusv, line 72: Doesn't this keep the quotes in case the string is not empty, but remove the quotes in case the string is empty?

juliusv:
@Nexucis Yeah, you will need to check whether there is a `MetricIdentifier` preceding the `LabelMatchers`, and translate that into a label matcher internally as well, before doing the linter checks.

src/lang-promql/parser/parser.ts
```diff
@@ -273,6 +264,45 @@ export class Parser {
...
+      this.state
+    );
+    // A Vector selector must contain at least one non-empty matcher to prevent
+    // implicit selection of all metrics (e.g. by a typo).
+    let empty = true;
+    let i = 0;
+    while (i < labelMatchers.length && empty) {
+      empty = labelMatchers[i].matchesEmpty();
+      i++;
+    }
```
juliusv, line 279: You can just replace the 6 lines above with:

```typescript
const empty = labelMatchers.every(lm => lm.matchesEmpty());
```

...and at that point you can probably just inline it.

src/lang-promql/parser/parser.ts
```diff
@@ -273,6 +264,45 @@ export class Parser {
...
+    if (labelMatchers.length > 0 && empty) {
```
juliusv, line 280: Once you include the metric name in the label matchers, I don't think it can ever happen that `labelMatchers.length` is 0, as then you would not end up in this code.

src/lang-promql/parser/matcher.ts
Nexucis: yeah it is. Maybe not so clever

juliusv: Guess you just want to do `matcher.value = value.slice(1,-1);`

Nexucis:
Thanks for your careful review :). I think I resolve all issue you spotted. I hope I didn't forget anything.

src/lang-promql/parser/parser.ts
```diff
@@ -269,31 +270,32 @@ export class Parser {
...
     // VectorSelector ( MetricIdentifier ( Identifier ) )
     // https://github.com/promlabs/lezer-promql/blob/71e2f9fa5ae6f5c5547d5738966cd2512e6b99a8/src/promql.grammar#L200
     const vectorSelectorNodeName = node.firstChild?.firstChild;
     if (vectorSelectorNodeName) {
       vectorSelectorName = this.state.sliceDoc(vectorSelectorNodeName.start, vectorSelectorNodeName.end);
     }
+    if (vectorSelectorName !== '') {
+      labelMatchers.push(new Matcher(EqlSingle, '__name__', vectorSelectorName));
+    }
+    console.log(`${this.tree}`);
```
juliusv, line 283: You forgot some debug statements :)

src/lang-promql/parser/parser.ts
```diff
@@ -273,6 +265,46 @@ export class Parser {
...
+    if (vectorSelectorName !== '') {
+      // In this case the last LabelMatcher is checking for the metric name
+      // set outside the braces. This checks if the name has already been set
+      // previously
+      let i = 0;
+      let metricNameSet = false;
+      while (i < labelMatchers.length - 1 && !metricNameSet) {
+        metricNameSet = labelMatchers[i].name === '__name__';
+        i++;
+      }
```
juliusv, line 301: Could just be:

```typescript
const metricNameSet = labelMatchers.some(lm => lm.name === '__name__');
```

src/lang-promql/parser/parser.ts
```diff
@@ -273,6 +265,46 @@ export class Parser {
...
+      if (metricNameSet) {
+        this.addDiagnostic(node, `metric name must not be set twice: ${vectorSelectorName} or ${labelMatchers[i - 1].value}`);
```
juliusv, line 303: Ah because of your `i` here... though you could then do:

```suggestion
        this.addDiagnostic(node, `metric name must not be set twice: ${vectorSelectorName} or ${labelMatchers.filter(lm -> lm.name === '__name__')[0].value}`);
```

juliusv: Even better, do this above:

```typescript
const labelMatcherMetricName = labelMatchers.find(lm => lm.name === '__name__');
```

That returns `undefined` in case there is no metric name.

juliusv: And then `labelMatcherMetricName.value` here.

juliusv:
Somehow with this, it still does not detect `{foo=""}` as wrong.

src/lang-promql/parser/parser.ts
Nexucis: If I'm doing that it will find the LabelMatcher I manually inject in the array.

juliusv: Maybe just do this first then, before extending the array :)

src/lang-promql/parser/matcher.ts
Nexucis: lol much better. Didn't know why I did it like this. Thanks !

Nexucis:
> Somehow with this, it still does not detect {foo=""} as wrong.

Huge mistake I made.  The name of the metric was finding like that: 

```typescript
const vectorSelectorNodeName = node.firstChild?.firstChild;
```
But it returns something when the expression is `{foo=""}`

It's fixed by specifying the exact path expecting like that:
```typescript
const vectorSelectorNodeName = walkThrough(node, MetricIdentifier, Identifier);
```

Now everything should be fine. Thanks again for your careful review. Hopefully you were there to catch the mistake made !

src/lang-promql/parser/parser.ts
Nexucis: yup I did it like you suggested, much cleaner. Thanks !

src/lang-promql/parser/parser.ts
Nexucis: mmm I should enable the rule in eslint to stop forgetting them

juliusv:
:+1: Nice :)

PR #41, opened 2020-09-06
Fix two linter bugs

juliusv:
The following expression should be legal, but was not:

    1 * 2

The following expression should not be legal, but was:

    metric_name * "string"


Nexucis (approved):
Nice ! Thanks

PR #43, opened 2020-09-08
Make snippet placeholders syntactically less problematic

juliusv:
The current placeholders wrapped in <> brackets would cause the linter to
believe that all kinds of things were wrong, because < and > are also
binary operators. Furthermore, it actually leads to a JS error when wanting
to autocomplete in the second field of the histogram_quantile() snippet,
but I couldn't track down why yet. It's definitely linter-related though,
as the JS error only happens with the linter enabled. This change at least
avoids that common situation for now.

### DIFF
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -59,16 +59,16 @@ const autocompleteNode: { [key: string]: AutoCompleteNode } = {
 
 const snippets: readonly SnippetSpec[] = [
   {
-    keyword: 'sum(rate(<input vector>[5m]))',
-    snippet: 'sum(rate(${<input vector>}[5m]))',
+    keyword: 'sum(rate(__input_vector__[5m]))',
+    snippet: 'sum(rate(${__input_vector__}[5m]))',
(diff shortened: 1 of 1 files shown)


juliusv:
@Nexucis Going to merge for now, please take a look after you're back from vacation :)

Nexucis:
Maybe escaping the characters would be a bit more elegant ? Not sure it would work better.

I'm fine with these changes. At least we should put a comment here to explain and say what should be done when a new snippet is creating. Just in order to avoid this situation again

juliusv:
@Nexucis Btw. I checked what CMN examples do for this, and they basically aren't showing snippets of this kind, they are just some descriptive text, and then the actual snippet code is just what gets inserted. We might want to change to something like that too, especially now that we can show extra info information for each item (which could be the snippet text, but you wouldn't match against it).

Nexucis:
Depending if we want to reproduce what the IDE like IntelliJ or VSC are doing @juliusv. If we don't then yeah it maybe make sense to hide the snippet code. I'm wondering what would be the matching text in that case.

Issue #48, opened 2020-09-19
Problems with LSP mode


dvkndn:
Following #46 , I created this issue to list the problems currently in the LSP mode (at least compare with the Prometheus mode).

Good: Prometheus
![Screenshot from 2020-09-20 03-57-25](https://user-images.githubusercontent.com/5953369/93689123-c1963680-faf5-11ea-9f69-859dba9b6d97.png)

Current bugs in LSP mode:
![Screenshot from 2020-09-20 03-57-44](https://user-images.githubusercontent.com/5953369/93689122-c0650980-faf5-11ea-960d-37ab10c83283.png)
1. [x] Matched text is not highlight
2. [x] Function's score is lower than metrics (actually I think the problem here is fuzzy score is not used?)
3. [ ] No snippet

Nexucis:
> Function's score is lower than metrics (actually I think the problem here is fuzzy score is not used?

It's on promql-langserver side. So filtering and scoring are not coming from cmn since it's on the server side to provide these features. @slrtbtfs can talk a bit more about it I believe :)

Nexucis:
In the PR https://github.com/prometheus-community/codemirror-promql/pull/8 that implements the autocompletion using LSP, there is a bunch of suggestions to improve it. Snippet was mentionned at this time.

There is certainly some code to do here, but also on promql-langserver side.

Hope it helps to clarify a bit the history of this mode / the situation

slrtbtfs:
The sorting order in the LSP mode is provided by the language server, in a way that functions appear first.

Maybe the code in this repo reverses that order, this would explain why the function appears last.

dvkndn:
Doesn't seem like this library uses the score returned by the language server: https://github.com/prometheus-community/codemirror-promql/blob/9e43058bd1341e8b52b0d8dcf1698a59e693b891/src/lang-promql/complete/lsp.ts#L74

Nexucis:
moving to cmn v12 should have fixed the first 2 points

dvkndn:
Yes I confirm the first 2 are fixed :D

Nexucis:
on this particular point, snippet should be implemented on server side. Nothing to do here ;).

dvkndn:
I think we can close this issue then :+1:

Issue #44, opened 2020-09-17
Text can become hidden behind tooltip


MacroPower:
The current positions of the diagnostic tooltips seem to be set relative to `start`.

I've noticed that as a result of this, when I have some issue spanning multiple lines, e.g.:
![image](https://user-images.githubusercontent.com/5648814/93522642-cabdb280-f8ff-11ea-99ba-b26ca7acd6f9.png)

The resulting tooltip makes it very difficult to edit the PromQL, as it completely covers part of the query and makes it impossible to click on the areas it covers. (Screenshot is after hovering over "...")
![image](https://user-images.githubusercontent.com/5648814/93522724-e923ae00-f8ff-11ea-9131-20c9391b6b50.png)

Example GIF from PromLens:
![output](https://user-images.githubusercontent.com/5648814/93528970-2d677c00-f909-11ea-8d7e-67bbfeb1823d.gif)

I think ideally this tooltip should either move with the cursor, or be set relative to `end` rather than start. On first glance, I'm not sure if there's any way to do this easily without making changes to codemirror.next, however.

dvkndn:
This is indeed CodeMirror's tooltip. To be specific, `codemirror-promql`'s lint uses `codemirror`'s lint, which uses `codemirror`'s tooltip with the following options:

https://github.com/codemirror/codemirror.next/blob/7aeef2544f9533b646183fe97394de1ead90434f/lint/src/lint.ts#L143-L152

> I think ideally this tooltip should either move with the cursor, or be set relative to end rather than start.

I think an easier approach is to set https://codemirror.net/6/docs/ref/#tooltip.Tooltip.above to `true`. This, however, still require CodeMirror's Lint module to allow us to set that option to Tooltip.

I think we should create an issue in CodeMirror asking for their opinion

dvkndn:
Fixed in upstream https://github.com/codemirror/codemirror.next/commit/d502554db22d48f9d6a4f1e03e3e8629dcc1b589 !

Issue #25, opened 2020-08-13
Snippets don't get applied correctly when selecting them via the mouse


juliusv:
When selecting a snippet such as `sum(rate(<input vector>[5m]))` with the mouse (not the keyboard), the snippet text is inserted verbatim, without `<input vector>` becoming a tabbable/completable field.

Nexucis:
maybe an upgrade of CMN to v0.10.0 will fix the issue.

Nexucis:
Problem is still there after the upgrade to CMN v0.10.0.
We should try with the vanilla app to see if the issue is only on our side or on cmn side too.

dvkndn:
Hi I did some more experiments, hope this helps:

1. The problem is still there (in codemirror-promql) after upgrade to CMN 0.12
2. The problem also exists on https://codemirror.net/6/ but it's a little bit different. You can see in the screenshot below that the placeholder part is selected but the editor is no longer focused. I believe this is a bug in CM

![Screenshot from 2020-09-20 03-42-34](https://user-images.githubusercontent.com/5953369/93688926-f903e380-faf3-11ea-92b7-5f2b75c0ee46.png)

This is different than codemirror-promql where there is no text selected at all (the editor still lost its focus though)

dvkndn:
Related: https://github.com/codemirror/codemirror.next/issues/287

juliusv:
@dvkndn Awesome, thanks for the testing and reporting it on the CMN side!

dvkndn:
Hey guys this is fixed upstream :D https://github.com/codemirror/codemirror.next/commit/c4ce6ee7bc6ac2b310bb63d74fb42cd90891cfe5 let me test it when CMN releases a new version

juliusv:
@dvkndn Nice! And since it's in the tooltip plugin and not our autocomplete plugin fork, it will hopefully just work for us as well. (it would still be nice if we didn't need our autocomplete plugin fork in the long term)

PR #50, opened 2020-09-30
move to cmn v0.12.0

### LINKED ISSUE #47: Do we still need the auto complete fork?
Afaik [the fork](https://github.com/Nexucis/codemirror-next-autocomplete/commits/master) is created for 2 primary reasons:
- Fuzzy search scoring https://github.com/codemirror/codemirror.next/issues/236
- Highlight matched string https://github.com/codemirror/codemirror.next/issues/234

Since both issues are addressed by codemirror's autocomplete module, do you think we can go back to the original module yet? Is there some plan for the fork?

Nexucis:
fix #47 
should partially solved some issue raised by #48 as well

### DIFF
--- a/src/app/app.ts
+++ b/src/app/app.ts
@@ -20,49 +20,9 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-import { EditorState, EditorView } from '@codemirror/next/basic-setup';
+import { basicSetup, EditorState, EditorView } from '@codemirror/next/basic-setup';
 import { PromQLExtension } from '../lang-promql';
-import { Extension } from '@codemirror/next/state';
-import { history, historyKeymap } from '@codemirror/next/history';
-import { highlightSpecialChars, keymap, multipleSelections } from '@codemirror/next/view';
-import { lineNumbers } from '@codemirror/next/gutter';
-import { foldGutter, foldKeymap } from '@codemirror/next/fold';
-import { bracketMatching } from '@codemirror/next/matchbrackets';
-import { closeBrackets, closeBracketsKeymap } from '@codemirror/next/closebrackets';
-import { autocomplete, autocompleteKeymap } from '@nexucis/codemirror-next-autocomplete';
-import { rectangularSelection } from '@codemirror/next/rectangular-selection';
-import { highlightActiveLine, highlightSelectionMatches } from '@codemirror/next/highlight-selection';
-import { defaultKeymap } from '@codemirror/next/commands';
-import { searchKeymap } from '@codemirror/next/search';
-import { commentKeymap } from '@codemirror/next/comment';
-import { gotoLineKeymap } from '@codemirror/next/goto-line';
-import { lintKeymap } from '@codemirror/next/lint';
 import { promQLHighlightMaterialTheme } from './theme';
-
-const basicSetup: Extension = [
-  lineNumbers(),
-  highlightSpecialChars(),
-  history(),
-  foldGutter(),
-  multipleSelections(),
-  bracketMatching(),
-  closeBrackets(),
-  autocomplete({ matchPre: '<b style="color: brown">', matchPost: '</b>' }),
-  rectangularSelection(),
-  highlightActiveLine(),
-  highlightSelectionMatches(),
-  keymap([
-    ...closeBracketsKeymap,
-    ...defaultKeymap,
-    ...searchKeymap,
-    ...historyKeymap,
-    ...foldKeymap,
-    ...commentKeymap,
-    ...gotoLineKeymap,
-    ...autocompleteKeymap,
-    ...lintKeymap,
-  ]),
-];
 
 const promqlExtension = new PromQLExtension();
 let editor: EditorView;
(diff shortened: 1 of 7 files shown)


Nexucis:
this PR should help to test if the issue #44 and #25 are fixed with cmn v0.13.0 once it is available too.

src/app/app.ts
```diff
@@ -20,50 +20,10 @@
...
-    ...searchKeymap,
-    ...historyKeymap,
-    ...foldKeymap,
-    ...commentKeymap,
-    ...gotoLineKeymap,
-    ...autocompleteKeymap,
-    ...lintKeymap,
-  ]),
-];
-
```
dvkndn, line 66: Just want to say that making the example simpler is great! Thank you

src/lang-promql/complete/hybrid.ts
```diff
@@ -59,25 +59,26 @@ const autocompleteNode: { [key: string]: AutoCompleteNode } = {
...
   },
 ];
 
 // parsedSnippets shouldn't be modified. It's only there to not parse everytime the above list of snippet
 const parsedSnippets: Completion[] = snippets.map((s) => ({
-  label: s.name || s.keyword,
-  original: s.name || s.keyword,
+  label: s.label,
   apply: snippet(s.snippet),
-  score: 0,
```
dvkndn, line 80: Please don't think this is a review, I just have a question. After this PR we still don't have scoring in this hybrid mode, right? I mean we haven't used CM's [boost](https://codemirror.net/6/docs/ref/#autocomplete.Completion.boost) option yet?

src/app/app.ts
Nexucis: :D the fork was just there to speed up the development and the integration of this lib into Prometheus and PromLens.
It was also a good way to test if cmn is working properly with another implementation of the autocompletion module.

But even if CMN is working as a module, it is still simpler to use the vanilla modules. It simplify so many things !

Nexucis: and you are really welcome :)

src/lang-promql/complete/hybrid.ts
Nexucis: actually the scoring is fully integrated in the autocomplete module. The `boost` is just a field to say "this particular label is really important". It's a way to influence a bit the internal scoring system.

For the moment I don't see with what condition we should use it, but you can be sure the list returned is using an internal score.

src/lang-promql/complete/hybrid.ts
```diff
@@ -59,25 +59,26 @@ const autocompleteNode: { [key: string]: AutoCompleteNode } = {
...
   {
-    keyword: 'histogram_quantile(__quantile__, sum by(le) (rate(__histogram_metric__[5m])))',
+    label: 'histogram_quantile(__quantile__, sum by(le) (rate(__histogram_metric__[5m])))',
+    type: 'function',
     snippet: 'histogram_quantile(${__quantile__}, sum by(le) (rate(${__histogram_metric__}[5m])))',
   },
   {
-    keyword: 'label_replace(__input_vector__, "__dst__", "__replacement__", "__src__", "__regex__")',
+    label: 'label_replace(__input_vector__, "__dst__", "__replacement__", "__src__", "__regex__")',
+    type: 'function',
```
juliusv, line 73: It seems this `type` is not propagated into the `parsedSnippets` and thus not used anywhere? Not sure it should be `function` though. Should we invent our own `snippet` type?

juliusv: And then if it's always `snippet` for snippets, I guess we can just hardcode it in the `parsedSnippets` below.

juliusv:
:+1: Super nice, thank you :)

Btw., the snippet selection by mouse now works for me, but the editor is unfocussed afterwards.

src/lang-promql/complete/hybrid.ts
Nexucis: It would be nice indeed to a have a dedicated icon for the snippet !

I will review the snippet management it seems I was a bit too fast here

Nexucis:
A bit weird the snippet is working since the patch is not yet released :o

src/lang-promql/complete/hybrid.ts
Nexucis: > And then if it's always snippet for snippets,

What do you mean by that ? Not sure I got your point actually :(.

Nexucis: I simplify the way to manage the snippet, hope you thought it like that too !

juliusv: Ah ok yeah, what you have now is fine too.

juliusv:
:+1:

src/lang-promql/complete/hybrid.ts
dvkndn: oh thank you. I've just went and check the autocomplete module and I saw the internal scoring system :+1:

src/lang-promql/complete/hybrid.ts
```diff
@@ -243,9 +237,7 @@ export class HybridComplete implements CompleteStrategy {
       }
     }
     if (includeSnippet) {
-      for (const s of parsedSnippets) {
-        options.push(s);
-      }
+      options.concat(snippets);
```
dvkndn, line 240: wait @Nexucis are you sure about this? I think `concat` doesn't change the `options` array at all

Nexucis: Ah yeah I forgot to get back the array returned. Thanks !

Nexucis: And yeah it's maybe better to follow your way. Should cost less to keep the array and to push the element.
Thanks for the suggestion

Nexucis: it's fixed. Thanks for the review @dvkndn :)

PR #52, opened 2020-10-01
enrich the completion with short details for the metric

Nexucis:
This PR proposed a way to add the help + the type of the metric in the completion list.

For the moment the help of the metric is not displayed. Just don't get why. I opened an issue on CMN side:
https://github.com/codemirror/codemirror.next/issues/300

And I don't know yet what short details we should put for the different function / aggregation method

info_not_displayed


src/lang-promql/complete/promql.terms.ts
```diff
@@ -1,54 +1,83 @@
...
+  { label: '>' },
+  { label: '<' },
+  { label: '<=' },
+  { label: '!=' },
+  { label: 'and' },
+  { label: 'or' },
+  { label: 'unless' },
+];
+export const binOpModifierTerms = [{ label: 'on' }, { label: 'ignoring' }, { label: 'group_left' }, { label: 'group_right' }];
 export const functionIdentifierTerms = [
```
dvkndn, line 20: > And I don't know yet what short details we should put for the different function / aggregation method

@Nexucis I think it would be great if we can show the first line of the function description from the doc, such as  https://github.com/prometheus/prometheus/blob/e17d18a0433a78b204dff7c3dbf33ff216471ed9/docs/querying/functions.md#changes

I don't know how @juliusv does that in PromLens but I think it's available in their "Explain" tab

src/lang-promql/client/prometheus.ts
```diff
@@ -218,6 +216,26 @@ export class HTTPPrometheusClient implements PrometheusClient {
     });
   }
 
+  metricMetadata(): Promise<Map<string, MetricMetadata[]>> {
+    if (this.cache.getMetricMetadata() && this.cache.getMetricMetadata().size > 0) {
+      return Promise.resolve(this.cache.getMetricMetadata());
+    }
+
+    return this.fetchAPI<Map<string, MetricMetadata[]>>(metricMetadataEndpoint)
```
juliusv, line 224: This is awesome, but I fear that this will be really expensive to fetch and keep in memory for large Prometheus server with up to 100k metrics or so. Not sure what to do though... maybe we would have to fetch it asynchronously only for the currently visible metrics, but there's probably no way to do that?

src/lang-promql/complete/promql.terms.ts
juliusv: I think a shorter, more dedicated manual text for this one-line description would be better. I have that in PromLens as well, but only in the form-based editor.

src/lang-promql/client/prometheus.ts
Nexucis: yeah that's true, this is a problem. I think the only way to address it is to change or add some new api in prometheus side. 
Not so bright idea regarding this point for the moment.
It would be nice if we could just get the metadata of the metric for a defined period of time like for others endpoint

src/lang-promql/complete/promql.terms.ts
Nexucis: when you are talking about `manual text`, are you talking about rewriting the description in order to have a shorter one ?

src/lang-promql/client/prometheus.ts
juliusv: One thing we could do is check how many metric names there are (we have those anyways) and have a limit above which we don't query the metadata anymore. Like 10k?

src/lang-promql/complete/promql.terms.ts
juliusv: Yes, correct. I have that for each function in PromLens, but of course at some point I have to think about what to OSS :laughing:

dvkndn: I believe so. The first line of most description in the doc are already good enough for example. Just need to rephrase some of them and we're good.

I'd be more than happy to contribute some manual work like that :D

juliusv: Oh well... if you're going to do it anyway. Here's a freebie:

https://gist.github.com/juliusv/1ab89524dcf53e0dd3e6ad313b681975

dvkndn: Oh guys turn out Grafana already did this https://github.com/grafana/grafana/blob/aae25c530873c0db51d987d1a1bfe84aeda0b3b1/public/app/plugins/datasource/prometheus/promql.ts#L16

Screenshot from 2020-10-01 22-18-14

juliusv: Ah right, we could also reuse that, as it has operators, etc. as well. Thanks!

Nexucis: yeah I guess so. I will just add the limit suggested by @juliusv and then merge the PR if there is nothing else to change / adapt

Nexucis: we have to wait CMN v0.13.0 to see the description of the function / metrics so it's ok if it's not done in this PR but in another.
The struct is there at least to handle it

Nexucis: note: in the gist of @juliusv it's missing the description of the aggregation operator (like sum / avg)

Nexucis: so feel free to open a PR @dvkndn whenever it fits you.

src/lang-promql/client/prometheus.ts
Nexucis: I put the limit in the completion code directly. I'm feeling it's better if this limit is driven by the completion and not by the client directly

juliusv:
:+1:

Nexucis:
Linked to #5

PR #53, opened 2020-10-02
upgrade lezer-promql to 0.12.0 and cmn 0.13.0

### LINKED ISSUE #25: Snippets don't get applied correctly when selecting them via the mouse

### LINKED ISSUE #44: Text can become hidden behind tooltip

Nexucis:
fix #44 
image

fix #25 
snippet-fixed

As you can see above, the help of the metric is now displayed

/cc @juliusv

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,38 +25,24 @@
       }
     },
     "@codemirror/next": {
-      "version": "0.12.0",
-      "resolved": "https://registry.npmjs.org/@codemirror/next/-/next-0.12.0.tgz",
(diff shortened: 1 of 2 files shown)


juliusv:
:+1:

PR #55, opened 2020-10-02
Add help info for Prom's built-in aggregation and function

dvkndn:
[As discussed](https://github.com/prometheus-community/codemirror-promql/pull/52#pullrequestreview-500252188), this PR adds help info for Prom's built-in aggregations and functions:

Screenshot from 2020-10-02 13-38-12

Screenshot from 2020-10-02 13-37-59

These description come from [Grafana's Prom plugin](https://github.com/grafana/grafana/blob/aae25c530873c0db51d987d1a1bfe84aeda0b3b1/public/app/plugins/datasource/prometheus/promql.ts#L16). Please feel free to suggest any change :D I think they are quite longer than @juliusv 's version.

Also [there is a bug](https://github.com/codemirror/codemirror.next/issues/302) in CodeMirror that the `max-width: 300px` of the Tooltip is not applied at all. I filed both issue and a PR to fix that in CMN repo.

### DIFF
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -43,7 +43,7 @@ import { walkBackward, walkThrough } from '../parser/path-finder';
 import { aggregateOpModifierTerms, aggregateOpTerms, binOpModifierTerms, binOpTerms, functionIdentifierTerms, matchOpTerms } from './promql.terms';
 import { Completion, CompletionContext, CompletionResult, snippet } from '@codemirror/next/autocomplete';
 
-interface AutoCompleteNode {
+export interface AutoCompleteNode {
   label: string;
   detail?: string;
   info?: string;
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -43,7 +43,7 @@ import { walkBackward, walkThrough } from '../parser/path-finder';
 import { aggregateOpModifierTerms, aggregateOpTerms, binOpModifierTerms, binOpTerms, functionIdentifierTerms, matchOpTerms } from './promql.terms';
 import { Completion, CompletionContext, CompletionResult, snippet } from '@codemirror/next/autocomplete';
 
-export interface AutoCompleteNode {
+interface AutoCompleteNode {
   label: string;
   detail?: string;
   info?: string;
--- a/src/lang-promql/complete/promql.terms.ts
+++ b/src/lang-promql/complete/promql.terms.ts
@@ -278,7 +278,7 @@ export const aggregateOpTerms = [
   {
     label: 'group',
     detail: 'aggregation',
-    info: 'All values in the resulting vector are 1',
+    info: 'Group series, while setting the sample value to 1'',
   },
   {
     label: 'max',
--- a/src/lang-promql/complete/promql.terms.ts
+++ b/src/lang-promql/complete/promql.terms.ts
@@ -278,7 +278,7 @@ export const aggregateOpTerms = [
   {
     label: 'group',
     detail: 'aggregation',
-    info: 'Group series, while setting the sample value to 1'',
+    info: 'Group series, while setting the sample value to 1',
   },
   {
     label: 'max',
--- a/src/lang-promql/complete/promql.terms.ts
+++ b/src/lang-promql/complete/promql.terms.ts
@@ -18,66 +18,254 @@ export const binOpTerms = [
 ];
 export const binOpModifierTerms = [{ label: 'on' }, { label: 'ignoring' }, { label: 'group_left' }, { label: 'group_right' }];
 export const functionIdentifierTerms = [
-  { label: 'abs' },
-  { label: 'absent' },
-  { label: 'absent_over_time' },
-  { label: 'avg_over_time' },
-  { label: 'ceil' },
-  { label: 'changes' },
-  { label: 'clamp_max' },
-  { label: 'clamp_min' },
-  { label: 'count_over_time' },
-  { label: 'days_in_month' },
-  { label: 'day_of_month' },
-  { label: 'day_of_week' },
-  { label: 'delta' },
-  { label: 'deriv' },
-  { label: 'exp' },
-  { label: 'floor' },
-  { label: 'histogram_quantile' },
-  { label: 'holt_winters' },
-  { label: 'hour' },
-  { label: 'idelta' },
-  { label: 'increase' },
-  { label: 'irate' },
-  { label: 'label_replace' },
-  { label: 'label_join' },
-  { label: 'ln' },
-  { label: 'log10' },
-  { label: 'log2' },
-  { label: 'max_over_time' },
-  { label: 'min_over_time' },
-  { label: 'minute' },
-  { label: 'month' },
-  { label: 'predict_linear' },
-  { label: 'quantile_over_time' },
-  { label: 'rate' },
-  { label: 'resets' },
-  { label: 'round' },
-  { label: 'scalar' },
-  { label: 'sort' },
-  { label: 'sort_desc' },
-  { label: 'sqrt' },
-  { label: 'stddev_over_time' },
-  { label: 'stdvar_over_time' },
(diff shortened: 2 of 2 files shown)


src/lang-promql/complete/promql.terms.ts
```diff
@@ -18,66 +18,254 @@ export const binOpTerms = [
...
-  { label: 'max' },
-  { label: 'min' },
-  { label: 'quantile' },
-  { label: 'stddev' },
-  { label: 'stdvar' },
-  { label: 'sum' },
-  { label: 'topk' },
+  {
+    label: 'avg',
+    info: 'Calculate the average over dimensions',
```
Nexucis, line 225: could be nice to add also the field `detail` with the value `aggregation`

src/lang-promql/complete/promql.terms.ts
```diff
@@ -18,66 +18,254 @@ export const binOpTerms = [
...
-  { label: 'stddev_over_time' },
-  { label: 'stdvar_over_time' },
-  { label: 'sum_over_time' },
-  { label: 'time' },
-  { label: 'timestamp' },
-  { label: 'vector' },
-  { label: 'year' },
+  {
+    label: 'abs',
+    info: 'Returns the input vector with all sample values converted to their absolute value.',
```
Nexucis, line 23: could be nice to add also the field `detail` with the value `function`

src/lang-promql/complete/promql.terms.ts
```diff
@@ -18,66 +18,254 @@ export const binOpTerms = [
...
+  },
+  {
+    label: 'absent',
+    info:
+      'Returns an empty vector if the vector passed to it has any elements and a 1-element vector with the value 1 if the vector passed to it has no elements. This is useful for alerting on when no time series exist for a given metric name and label combination.',
```
Nexucis, line 28: on this point, I think the simpler should be to use what @juliusv shared. At least for the info that are quite long. But maybe Julius will have a different opinion.
It wasn't super clear for me if we should take as is what Grafana is proposing

src/lang-promql/complete/promql.terms.ts
```diff
@@ -18,66 +18,254 @@ export const binOpTerms = [
...
+  {
+    label: 'count',
+    info: 'Count number of elements in the vector',
+  },
+  {
+    label: 'count_values',
+    info: 'Count number of elements with the same value',
+  },
+  {
+    label: 'group',
```
Nexucis, line 240: `info` is missing here. Is it because you didn't find any documentation ?

Nexucis (commented):
thanks for the PR :).
I added some minor comments. Nothing too serious.

src/lang-promql/complete/promql.terms.ts
dvkndn: ah yes I ran a script to generate this from Grafana but they don't have a description for this one. I'll add one manually from Prometheus

src/lang-promql/complete/promql.terms.ts
dvkndn: Fixed: 8ec7de3741e9e6488400e7320eb56c74e2badd94

Screenshot from 2020-10-02 14-02-03

src/lang-promql/complete/promql.terms.ts
dvkndn: Fixed https://github.com/prometheus-community/codemirror-promql/pull/55#discussion_r498647747

src/lang-promql/complete/promql.terms.ts
dvkndn: After reading and trying both I strongly believe @juliusv 's version is much better as "help". It's simpler and has the right length. I made a commit to use Julius's version 7f23ed9ee5932498dc4d55af4b3628001c4fffb2 If he accepts this then we are good, or else I will reverse back to the Grafana and try to shorten them

src/lang-promql/complete/promql.terms.ts
dvkndn: Added f77b4954f8f560151b19345fb826410ea388e0ba

dvkndn:
Small update that the [max-width fix](https://github.com/codemirror/codemirror.next/pull/303) is merged :D

src/lang-promql/complete/promql.terms.ts
```diff
@@ -17,67 +19,251 @@ export const binOpTerms = [
...
+    info: 'All values in the resulting vector are 1',
```
juliusv, line 234: While this is true, it misses the main purpose of the `group` aggregator, which is not the sample value, but the grouping. How about:


```suggestion
    info: 'Group series, while setting the sample value to 1'',
```

src/lang-promql/complete/promql.terms.ts
```diff
@@ -17,67 +19,251 @@ export const binOpTerms = [
...
+  {
+    label: 'sum',
+    info: 'Calculate sum over dimensions',
+  },
+  {
+    label: 'topk',
+    info: 'Largest k elements by sample value',
+  },
 ];
+aggregateOpTerms.forEach((term) => {
```
juliusv, line 265: It would be nicer to not mutate the existing array, but to just assign the final one like:

```typescript
export const aggregateOpTerms: AutoCompleteNode[] = [
  // ...
].map((term) => {...term, detail: 'aggregation'});
```

Nexucis: I would more suggest to duplicate the label detail everywhere. It's stupid but simpler and clearer

juliusv: Also fine with me.

src/lang-promql/complete/promql.terms.ts
```diff
@@ -17,67 +19,251 @@ export const binOpTerms = [
...
-  { label: 'group' },
-  { label: 'max' },
-  { label: 'min' },
-  { label: 'quantile' },
-  { label: 'stddev' },
-  { label: 'stdvar' },
-  { label: 'sum' },
-  { label: 'topk' },
+functionIdentifierTerms.forEach((term) => {
+  term.detail = 'function';
```
juliusv, line 213: Similar note here as with the other list.

src/lang-promql/complete/hybrid.ts
```diff
@@ -43,7 +43,7 @@ import { walkBackward, walkThrough } from '../parser/path-finder';
 import { aggregateOpModifierTerms, aggregateOpTerms, binOpModifierTerms, binOpTerms, functionIdentifierTerms, matchOpTerms } from './promql.terms';
 import { Completion, CompletionContext, CompletionResult, snippet } from '@codemirror/next/autocomplete';
 
-interface AutoCompleteNode {
+export interface AutoCompleteNode {
```
Nexucis, line 46: I don't see the point of exporting this term. I would prefer that the file promql.term.ts doesn't depend to anything in order to avoid a nasty circular dependency

dvkndn: It's because of the code that modifies the terms array to insert the "detail" after the array is created (so TypeScript requires the type)

Anyway, it's no longer necessary as I will put the detail directly like your suggestion :+1:

src/lang-promql/complete/promql.terms.ts
dvkndn: Thank you. Fixed in a5aa063ad5a613cd23c50be6a0cc9e795ccca57b

src/lang-promql/complete/promql.terms.ts
dvkndn: Fixed https://github.com/prometheus-community/codemirror-promql/pull/55#discussion_r498716990

src/lang-promql/complete/hybrid.ts
dvkndn: Removed in https://github.com/prometheus-community/codemirror-promql/pull/55#discussion_r498716990

Nexucis:
@dvkndn looks like the build is broken :(. Could you try to fix it please ?

Nexucis:
Awesome ! Thanks for your contribution !

PR #56, opened 2020-10-02
Fully rewriting Hybrid Autocomplete with associated unit test

Nexucis:
This PR is proposing a new way to manage the autocompletion.

First the code is going to analyze what kind of autocompletion it should do using an enum.
Then the result will be used to then retrieved the required information.

**Note**: it's a bit more verbose, but definitely easier to test and I believe it will be easier to improve it.
**Note2**: it will certainly make some conflict with the PR #55 

related to #5

### DIFF
--- a/src/lang-promql/complete/hybrid.test.ts
+++ b/src/lang-promql/complete/hybrid.test.ts
@@ -33,6 +33,30 @@ describe('analyzeCompletion test', () => {
       expectedContext: [{ kind: ContextKind.MetricName }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation }],
     },
     {
+      title: 'metric/function/aggregation autocompletion',
+      expr: 'sum()',
+      pos: 4,
+      expectedContext: [{ kind: ContextKind.MetricName }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation }],
+    },
+    {
+      title: 'metric/function/aggregation autocompletion 2',
+      expr: 'sum(rat)',
+      pos: 7,
+      expectedContext: [{ kind: ContextKind.MetricName }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation }],
+    },
+    {
+      title: 'metric/function/aggregation autocompletion 3',
+      expr: 'sum(rate())',
+      pos: 9,
+      expectedContext: [{ kind: ContextKind.MetricName }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation }],
+    },
+    {
+      title: 'metric/function/aggregation autocompletion 4',
+      expr: 'sum(rate(my_))',
+      pos: 12,
+      expectedContext: [{ kind: ContextKind.MetricName }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation }],
+    },
+    {
       title: 'autocomplete binOp modifier or metric',
       expr: 'metric_name / ignor',
       pos: 19,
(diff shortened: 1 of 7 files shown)


src/lang-promql/complete/hybrid.test.ts
```diff
@@ -0,0 +1,134 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import chai from 'chai';
+import { ContextKind, HybridComplete } from './hybrid';
+import { createEditorState } from '../../test/utils';
+
+describe('analyzeCompletion test', () => {
+  const testSuites = [
```
juliusv, line 28: A suite is a whole collection of test cases, sounds like this should be called `testCases`?

src/lang-promql/complete/hybrid.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    },
+    {
+      title: 'autocomplete matchOp',
+      expr: 'go{instance=""}',
+      pos: 12, // cursor is after the 'equal'
+      expectedContext: [{ kind: ContextKind.MatchOp }],
+    },
+    {
+      title: 'autocomplete matchOp 2',
+      expr: 'metric_name{labelName!}',
```
juliusv, line 120: This isn't working for me... if I type `metric_name{labelName!<cursor>}` then it doesn't autocomplete `!=` for me. It only autocompletes those match ops if the first character is already a valid one, like `=` completing to `=` or `=~`.

src/lang-promql/complete/hybrid.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    },
+    {
+      title: 'autocomplete aggregate operation modifier 2',
+      expr: 'sum b',
+      pos: 5, // cursor is after 'b'
+      expectedContext: [{ kind: ContextKind.AggregateOpModifier }, { kind: ContextKind.BinOp }],
+    },
+    {
+      title: 'autocomplete binOp',
+      expr: 'metric_name unle',
```
juliusv, line 108: Could you add a case for `metric_name !<cursor>`? It should autocomplete `!=`, but it doesn't. Similar for the other ops, it only autocompletes when the typed token is already a valid op.

juliusv: (it worked before btw., same for the other comment)

Nexucis: Mmm that's weird. Does it still work on the master branch ?
I'm wondering if it's me that breaks everything or it's coming with cmn v0.13.0

juliusv: Yeah, it still seems to work on master!

Nexucis: :'( needs more test then hahaha ! Thanks for the check :)

src/lang-promql/complete/hybrid.test.ts
Nexucis: I finally found why it doesn't work. That's because the `start` position for the completion is wrong for this case.

In the previous case it was `tree.lastChild.start` and now it is `tree.start`.

I will add some unit test to cover the calculation of the `start` position. It's a bit critical step that condition if the autocompletion will work or not even if we have properly determinated the list of the autocompletion

src/lang-promql/complete/hybrid.test.ts
Nexucis: yup indeed, this cases were missing. It's covered now

src/lang-promql/complete/hybrid.test.ts
Nexucis: fixed

src/lang-promql/complete/hybrid.test.ts
Nexucis: fixed

juliusv:
:+1: besides the discussed issues

Nexucis:
thanks for the review and the constant test @juliusv !

PR #60, opened 2020-10-06
Improve linter with vectorMatching analyzing

Nexucis:
this PR finalized the linter. Hopefully, it should cover every possible error in promQL !

### DIFF
--- a/src/lang-promql/parser/parser.ts
+++ b/src/lang-promql/parser/parser.ts
@@ -59,7 +59,7 @@ import {
   VectorSelector,
 } from 'lezer-promql';
 import { containsAtLeastOneChild, retrieveAllRecursiveNodes, walkThrough } from './path-finder';
-import { getFunction, getType, ValueType } from './type';
+import { buildVectorMatching, getFunction, getType, ValueType, VectorMatchCardinality } from './type';
 import { buildLabelMatchers, Matcher } from './matcher';
 import { EditorState } from '@codemirror/next/state';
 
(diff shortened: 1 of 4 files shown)


src/lang-promql/parser/parser.ts
```diff
@@ -209,14 +209,39 @@ export class Parser {
...
     }
+
+    if ((lt !== ValueType.vector || rt !== ValueType.vector) && vectorMatching !== null) {
+      if (vectorMatching.matchingLabels.length > 0) {
+        this.addDiagnostic(node, 'vector matching only allowed between instant vectors');
+      }
+    } else {
+      if (isSetOperator) {
+        if (vectorMatching?.card === VectorMatchCardinality.CardOneToMany || vectorMatching?.card === VectorMatchCardinality.CardManyToOne) {
+          this.addDiagnostic(node, 'no grouping allowed for this kind of operation');
```
juliusv, line 238: I think we could be even clearer here (and probably also improve those messages in Prometheus / LSP):

```suggestion
          this.addDiagnostic(node, 'no grouping allowed for set operations');
```

juliusv:
:+1: Amazing, thanks!

src/lang-promql/parser/parser.ts
Nexucis: In prometheus it's a bit more clearer since it has in the message the name of the set operator.
But maybe you are thinking about something else ?

Nexucis: fixed, thanks for the review :)

juliusv: Ah yeah, that would also be a good option. Either including the name of the operator or just saying it's a set operator would be clearer than just "this kind of operation".

Nexucis: Currently I don't have the name of the operator, I just know it's a setOperator. I would suggest to do it a bit later once CMN improves the way to iterate other the tree.

I think there is a lot of refactoring to improve the code in the linter, to make it clearer and simpler once the iteration of the tree is fixed

PR #63, opened 2020-10-08
compute a bit smarter the metadata when there are multiple definitions for a given metric

Nexucis:


src/lang-promql/complete/hybrid.ts
```diff
@@ -486,7 +486,19 @@ export class HybridComplete implements CompleteStrategy {
...
+                  if (node.detail !== '' && node.detail !== m.type) {
+                    node.detail = 'unknown';
+                    node.info = 'multiple different definitions for this metric';
+                  } else if (node.detail === '') {
+                    node.detail = m.type;
+                    node.info = m.help;
+                  }
+                  // determinate what to do with the help
+                  if (node.info !== m.help) {
+                    node.info = 'multiple different definitions for this metric';
```
juliusv, line 499: Can you not put this in the first `if` condition above? Like:

```typescript
                for (const m of metadata) {
                  if (node.detail !== '' && (node.detail !== m.type || node.info !== m.help)) {
                    node.detail = 'unknown';
                    node.info = 'multiple different definitions for this metric';
                  } else if (node.detail === '') {
                    node.detail = m.type;
                    node.info = m.help;
                  }
```

Btw., it could be a bit clearer if you switched the two cases around and started with the case where there is no info yet:

```typescript
                for (const m of metadata) {
                  if (node.detail === '') {
                    node.detail = m.type;
                    node.info = m.help;
                  } else if (node.detail !== '' && (node.detail !== m.type || node.info !== m.help)) {
                    node.detail = 'unknown';
                    node.info = 'multiple different definitions for this metric';
                  } 
```

Nexucis: if you consider you have the following metadata:

```json
{
  "go_threads": [
    {
       "type:": "gauge",
       "help": "Number of OS threads created."
    },
    {
       "type:": "gauge",
       "help": "Number of OS threads created"
    }
  ]
}
```

The help are different but the type are not. So the code I'm proposing here would result as:
```
node.detail = "gauge"
node.info = "multiple different definitions for this metric"
```

And with the code you are proposing, result will be:

```
node.detail = "unknown"
node.info = "multiple different definitions for this metric"
```

A bit shame from my point of view since, the type of the metrics is the same and only the help is changing.

juliusv: Ah yeah, good point, then keep it :) :+1:

Nexucis: for your 2nd point, actually I did it on purpose to avoid to pass for most of the case in the two `if`. But I guess it doesn't really matter since the list of the metadata should be quite short.
So I will change it like you propose.

Nexucis: I change it, but with a different way, I found it clearer. Tell me if you think the same ! ;)

juliusv:
:+1: Thanks :)

PR #64, opened 2020-10-10
autocomplete duration

Nexucis:
This PR adds the autocompletion of the duration for the MatrixSelector.

It's also adding a `filter` list in the context that is used for the moment to remove some labelValue already used in the vectorSelector.

That's what illustrate the gif below. It shows that when the user want to use again the label 'instance' there is no value autocompleted since it has been already used before.

duration

I also custom a bit the theme in the app just for the fun.

### DIFF
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -46,6 +46,7 @@ import {
   Lss,
   Lte,
   MatchOp,
+  MatrixSelector,
   MetricIdentifier,
   Mod,
   Mul,
@@ -132,7 +133,7 @@ function getMetricNameInVectorSelector(tree: Subtree, state: EditorState): strin
   return state.sliceDoc(currentNode.start, currentNode.end);
 }
 
-function arrayToCompletionResult(data: AutoCompleteNodes[], from: number, to: number, includeSnippet = false): CompletionResult {
+function arrayToCompletionResult(data: AutoCompleteNodes[], from: number, to: number, includeSnippet = false, span = true): CompletionResult {
   const options: Completion[] = [];
 
   for (const completionList of data) {
@@ -152,7 +153,7 @@ function arrayToCompletionResult(data: AutoCompleteNodes[], from: number, to: nu
     from: from,
     to: to,
     options: options,
-    span: /^[a-zA-Z0-9_:]+$/,
+    span: span ? /^[a-zA-Z0-9_:]+$/ : undefined,
   } as CompletionResult;
 }
 
@@ -181,7 +182,11 @@ export function computeStartCompletePosition(node: Subtree, pos: number): number
     // if start would be `node.start`, then at the end it would try to autocomplete a string that starts with labelName! and not by !
     // `!` is contain in the error node so in the lastChild in this situation
     start = node.lastChild.start;
-  } else if (node.type.id === OffsetExpr || (node.type.id === 0 && node.parent?.type.id === OffsetExpr)) {
+  } else if (
+    node.type.id === OffsetExpr ||
+    (node.type.id === MatrixSelector && containsAtLeastOneChild(node, 0)) ||
+    (node.type.id === 0 && (node.parent?.type.id === OffsetExpr || node.parent?.type.id === MatrixSelector))
+  ) {
     start = pos;
   }
   return start;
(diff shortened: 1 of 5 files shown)


src/lang-promql/complete/hybrid.ts
```diff
@@ -331,12 +346,34 @@ export function analyzeCompletion(state: EditorState, node: Subtree): Context[]
...
         // then find the metricName if it exists
         const metricName = getMetricNameInVectorSelector(node, state);
-        result.push({ kind: ContextKind.LabelValue, metricName: metricName, labelName: labelName });
+        const context = {
+          kind: ContextKind.LabelValue,
+          metricName: metricName,
+          labelName: labelName,
+          filters: [] as string[],
+        };
+        // In addition, it is interesting to know if there is other labelMatcher in the current VectorSelector.
```
juliusv, line 360: I wonder if it's worth adding this extra complexity, as the number of possible label names on a given metric is already small anyway, and at least in principle, it is legal to use the same label name multiple times, with different matching conditions. For example, it is not common, but you could do something like `foo{fruit!="banana",fruit!="apple"}` (of course you could also shorten that with a negative regex selector).

src/lang-promql/complete/hybrid.ts
```diff
@@ -427,6 +473,7 @@ export class HybridComplete implements CompleteStrategy {
           });
           break;
         case ContextKind.Duration:
+          span = false;
```
juliusv, line 476: Maybe not super important right now, but we might also want a custom span regex for e.g. label values, which can be arbitrary strings?

juliusv:
Two notes:

* Should we really already autocomplete before any digit has been typed? It seems like that would be suggesting something that is syntactically wrong.
* The alphabetical ordering of the units in the autocomplete might not be the most useful one. Sorting by the magnitude of the unit might be better, but that would put `ms` before `m` and `m` is the unit that people will want in 90% of cases. Hmm... wondering if we should just omit `ms` then or something, I'm unsure.

src/lang-promql/complete/hybrid.ts
Nexucis: Actually it's only for the labelValue not the labelName.
If I'm following your example, then when you would start to autocomplete the 2nd 'fruit' it won't return 'banana' because it's already used for the labelName fruit

Nexucis:
I think it's not possible to change the way it is sorted unfortunately. It's completely on cmn side

src/lang-promql/complete/hybrid.ts
Nexucis: That's possible I'm just not sure about the value here

src/lang-promql/complete/hybrid.ts
Nexucis: If you look at the gif in the PR description you will see that the label `instance` is still autocomplete but the value is not for the 2nd time

juliusv: Ah sorry, I didn't notice that it was doing it for the label value, not the name. In that case though it seems even more exotic because almost nobody repeats the same label name twice in a selector, so it adds a bit of code complexity for like a 1% use case.

juliusv:
Hmm ok. Then let's leave the sorting like it is for now. Grafana takes a different approach where they have some hardcoded examples like `1m` / `5m` / 15m` that they autocomplete, but not sure if that's much better.

src/lang-promql/complete/hybrid.ts
juliusv: Yeah I would also have to look deeper, maybe it would just be `.*`? Anyway, not important for this PR.

src/lang-promql/complete/hybrid.ts
Nexucis: mmm yeah you are certainly right. I will remove the code then

src/lang-promql/complete/hybrid.ts
Nexucis: mmm ok, then I will let you take a look once you have the time. I think I missed something here.

Nexucis:
> Should we really already autocomplete before any digit has been typed? It seems like that would be suggesting something that is syntactically wrong.

That's a good question. I think a longer usage of this autocompletion will tell us if it's annoying or not. Or would you prefer to remove it directly ?

src/lang-promql/complete/hybrid.ts
Nexucis: removed

juliusv:
> That's a good question. I think a longer usage of this autocompletion will tell us if it's annoying or not. Or would you prefer to remove it directly ?

I think I'd prefer to only autocomplete things that are syntactically legal, so I'd be for only autocompleting the unit only when there is at least one digit already.

Nexucis:
ok :), I changed it. So it won't autocomplete the duration if you have this expression : `my_metric[]`, but it will if you have `my_metric[85]`

juliusv:
Hmm, now when I try it on GitPod, it only autocompletes in some situations, not others, it's a bit weird:

autocomplete

Nexucis:
Should be good now @juliusv, I didn't cleanup everything when I removed the code that was adding the autocompletion of the duration when brackets are empty. It was coming from that mistake.

juliusv:
:+1: Nice, seems to work great now :)

Issue #69, opened 2020-10-15
Incorrect linter errors when whitespace around binops is missing


juliusv:
Try `a+b`, it gives a linter error for `b`. `a+ b` works though. Autocomplete seems to work in all cases, so not sure it's a lezer grammar bug or just a linter bug.

Nexucis:
it's a problem with the method `walkThrough` that doesn't find the node corresponding to `b`. 
It's used here: https://github.com/prometheus-community/codemirror-promql/blob/master/src/lang-promql/parser/parser.ts#L300

It's linked to this issue and comment: https://github.com/codemirror/codemirror.next/issues/270#issuecomment-704420758

For this particular case, it doesn't really to have the `Subtree` corresponding to `b`, it's just need to have the start/end position. So it's still possible to create another method that returns these information instead of trying to find the node and failed to find it

PR #72, opened 2020-10-16
Improve walkThrough() node resolution by using end position

juliusv:
This correctly finds the metric name "b" in "a+b", where previously it
did not.

Fixes https://github.com/prometheus-community/codemirror-promql/issues/69

### DIFF
--- a/src/lang-promql/parser/parser.test.ts
+++ b/src/lang-promql/parser/parser.test.ts
@@ -115,6 +115,11 @@ describe('Scalars and scalar-to-scalar operations', () => {
       expectedDiag: [] as Diagnostic[],
     },
     {
+      expr: 'foo*bar',
+      expectedValueType: ValueType.vector,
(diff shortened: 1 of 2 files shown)


src/lang-promql/parser/parser.test.ts
```diff
@@ -114,6 +114,11 @@ describe('Scalars and scalar-to-scalar operations', () => {
       expectedValueType: ValueType.vector,
       expectedDiag: [] as Diagnostic[],
     },
+    {
+      expr: 'foo*bar',
```
Nexucis, line 118: Would be worth add these tests :

`a +b`
`a+b`
`a+ b`
`a==b`

The first 3 tests are to check the space effect, the last one for binOps expression that are long than 1 char

juliusv: Yup, added.

Nexucis (approved):
👍 thank you !

Issue #70, opened 2020-10-15
Processing of fetched autocomplete results blocks browser


juliusv:
When I autocomplete something like this for a metric with many series:

```
container_cpu_usage_seconds_total{inst<cursor>}
```

...the typing is fast initially, but once the autocomplete fetch request has completed, the processing of the huge result in JS takes multiple seconds (like ~5s or so in my case) and hangs the whole browser for that period of time since JS is single-threaded. I wonder if there's something to make that better a) in this specific case by optimizing it, b) in general, by changing the processing architecture in some way.

Nexucis:
I think we need some benchmark test to see exactly what to optimize

PR #73, opened 2020-10-16
Speed up label value caching by using Set

juliusv:
For a large metric, this speeds up the setting of the cache from 10 seconds
to 160ms.

Fixes https://github.com/prometheus-community/codemirror-promql/issues/70


Nexucis (approved):
Awesome ! 👍

PR #76, opened 2020-10-18
fix autocomplete binary operator after aggregration

### LINKED ISSUE #75: Set operators don't autocomplete after aggregations
This autocompletes `and` correctly:

    foo an<cursor>

This as well:

    rate(foo[5m]) an<cursor>

But this does not:

    sum(rate(foo[5m])) an<cursor>

However, other binops complete correctly in the same position after an aggregation.

Nexucis:
fix #75


juliusv:
:+1: Haha oops, I just made pretty much the same change without seeing yours first: https://github.com/prometheus-community/codemirror-promql/commit/9ab49ce8a76a36dc6b4760834dc02e8456d936bc

Yours wins ;)

PR #79, opened 2020-10-28
Fix the way to retrieve the parameter in the function

### LINKED ISSUE #78: Linter should ignore comments
Comments seem to influence linting behavior and introduce linter warnings where there should be none.

Example: https://promlens.com/?l=E7B5fVU6KQ3

Which looks like this:

comments-linter-error

The linter error in this case is "expected 2 argument(s) in call to histogram_quantile, got 0".

When removing the comments, everything is fine.

Nexucis:
Before this PR, it was assuming that the recursive node was the firstChild. But as you can put a comment literally anywhere you would like, then the first child can be a comment.

So instead we just have to search where is the recursive node in the child (using the method `walkThrough` because that's the easiest way for the moment. It should be deeply reviewed when using the new lezer version (v0.12.0))

fix #78 

Note: I guess this problem should appear in the Binary operation as well and more generally everywhere the code is assuming that a node is at the first place in the child.

### DIFF
--- a/src/lang-promql/parser/parser.test.ts
+++ b/src/lang-promql/parser/parser.test.ts
@@ -26,7 +26,7 @@ import { ValueType } from './type';
 import { Diagnostic } from '@codemirror/next/lint';
 import { createEditorState } from '../../test/utils';
 
-describe('Scalars and scalar-to-scalar operations', () => {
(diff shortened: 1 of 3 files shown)


juliusv:
:+1: Thanks!

PR #80, opened 2020-11-03
Moving to lezer v0.12.0 through codemirror.next v0.14.0

Nexucis:
This is a deep review of how we should iterate over the tree.

Note: as a side effect, it looks like the resolution of the tree is much more accurate and make more sense according to different weird case we had to patch with ugly code. ( like the autocompletion of the `MatchOp` )

And this is really cool we have many tests. Otherwise it would have been really hard to find what was broken after making this move.

### DIFF
--- a/src/lang-promql/parser/parser.ts
+++ b/src/lang-promql/parser/parser.ts
@@ -21,7 +21,7 @@
 // SOFTWARE.
 
 import { Diagnostic } from '@codemirror/next/lint';
-import { Subtree, Tree } from 'lezer-tree';
+import { SyntaxNode, Tree } from 'lezer-tree';
 import {
   AggregateExpr,
   And,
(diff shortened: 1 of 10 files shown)


Nexucis:
and perhaps there is a way to improve it by using a bit more the feature from lezer, but I don't have better idea for the moment. At least if someone want to do it, it's now possible.

src/lang-promql/parser/parser.ts
```diff
@@ -83,33 +83,28 @@ export class Parser {
...
     // So to be able to iterate other the node relative to the promql node, we have to get the first child at the beginning
-    this.checkAST(this.tree.firstChild);
+    this.checkAST(this.tree.topNode.firstChild);
     this.diagnoseAllErrorNodes();
   }
 
   private diagnoseAllErrorNodes() {
-    this.tree.iterate({
-      enter: (type, start, end) => {
-        // usually there is an error node at the end of the expression when user is typing
```
juliusv, line 93: Could you keep this comment? Makes it much easier to understand the purpose :)

Nexucis: done

juliusv:
:+1: Awesome :)

juliusv:
(I also played around a bit with it and couldn't find any breakage yet)

PR #81, opened 2020-11-04
 add a parameter for the limit of the metrics

Nexucis:


src/lang-promql/complete/hybrid.ts
```diff
@@ -358,9 +358,11 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
 // HybridComplete provides a full completion result with or without a remote prometheus.
 export class HybridComplete implements CompleteStrategy {
   private readonly prometheusClient: PrometheusClient | undefined;
+  private readonly maxLimitMetricsMetadata: number;
 
-  constructor(prometheusClient?: PrometheusClient) {
+  constructor(prometheusClient?: PrometheusClient, maxLimitMetricsMetadata = 10000) {
```
juliusv, line 363: `max` and `limit` seems redundant. Maybe just `maxMetricsMetadata`?

src/lang-promql/promql.ts
```diff
@@ -57,6 +57,8 @@ export const promQLSyntax = LezerSyntax.define(
   }
 );
 
+const simplePromQLSyntax = LezerSyntax.define(parser);
```
juliusv, line 60: IMO the node props added above should always be added. They are necessary for correct highlighting, but they are separate from highlighting. I noticed the highlighter is already an external separate extension added in the example app (https://github.com/prometheus-community/codemirror-promql/blob/17cecae37707548e5bc7fb9906ef0d7f97e351e7/src/app/theme.ts#L6-L26).

When you were asking the question on chat, I thought you were considering moving that highlighter into the PromQL extension and making that optionally enabled.

Nexucis: Well that's a way to deactivate the highlight :).

Once it's removed, then of course the highlighter is not working anymore. That's a bit the purpose. But if you activate it then you have the props added.

I know it's necessary for a correct highlightning, but if the purpose is to not highlight, then this is an (easy) solution to achieve, without altering the way codemirror is configured

src/lang-promql/complete/hybrid.ts
Nexucis: Ah yeah sounds good. I wasn't super confortable with the name of this variable

src/lang-promql/promql.ts
juliusv: Ok, while that technically achieves an outcome where the user will not see highlighting, if the highlighter is still separate anyway, then it's IMO much better to just keep it that way and let the user control highlighting by either including a highlighter or not. I don't think we should be removing this node metadata...

Nexucis: ok :) It wasn't a so good idea afterall :D. I will remove it then. I think I didn't get your point too in the tchat and didn't know it was already achievable with CMN natively.

src/lang-promql/complete/hybrid.ts
Nexucis: done

juliusv:
:+1:

PR #82, opened 2020-11-16
Add a way to customize the completion result

Nexucis:
This PR provides a way to inject a custom completion. It won't override the current completion but will enrich it instead.
It will be possible to trigger the custom completion for a certain kind of completion (like only when the system will autocomplete the aggregation)

related to https://github.com/prometheus/prometheus/pull/8174

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,47 @@ In case you would like to deactivate the linter and/or the autocompletion it's s
 const promQL = new PromQLExtension().activateLinter(false).activateCompletion(false) // here the linter and the autocomplete are deactivated
 ```
 
+### maxMetricsMetadata
+
+maxMetricsMetadata is the maximum limit of the number of metrics in Prometheus. Under this limit, it allows the completion to get the metadata of the metrics.
+
+By default, the limit is 10 000 metrics. 
+
+Use it cautiously. A high value of this limit can cause a crash of your browser due to too many data fetched.
+
+```typescript
+const promQL = new PromQLExtension().setComplete({maxMetricsMetadata: 10000})
+```
+
+### enricher
+
+enricher is a function that will allow user to enrich the current completion by adding a custom one. It takes two parameters:
+
+* `trigger` should be used to decide whenever you would like to trigger your custom completion.
+  It's better to not trigger the same completion for multiple different ContextKind.
+  For example, the autocompletion of the metricName / the function / the aggregation happens at the same time.
+  So if you want to trigger your custom completion for metricName / function / aggregation, you should just choose to trigger it for the function for example.
+  Otherwise, you will end up to have the same completion result multiple time.
+
+* `result` is the current result of the completion. Usually you don't want to override it but instead to concat your own completion with this one.
+
+```typescript
+import { Completion } from '@codemirror/next/autocomplete';
+import { ContextKind } from 'codemirror-promql';
+
+function myCustomEnricher(trigger: ContextKind, result: Completion[]): Completion[] | Promise<Completion[]> {
+  switch (trigger) {
+      case ContextKind.Aggregation:
+        // custom completion
+        // ...
+        // return result.concat( myCustomCompletionArray )
+      default:
+        return result;
+    }
+}
+const promQL = new PromQLExtension().setComplete({enricher: myCustomEnricher})
+```
+
 ### Connect the autocompletion extension to a remote Prometheus server
 Connecting the autocompletion extension to a remote Prometheus server will provide autocompletion of metric names, label names, and label values.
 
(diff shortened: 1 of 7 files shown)


src/lang-promql/complete/hybrid.ts
```diff
@@ -370,10 +394,12 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
 export class HybridComplete implements CompleteStrategy {
   private readonly prometheusClient: PrometheusClient | undefined;
   private readonly maxMetricsMetadata: number;
+  private readonly enricher: EnrichCompletionHandler;
 
-  constructor(prometheusClient?: PrometheusClient, maxMetricsMetadata = 10000) {
+  constructor(prometheusClient?: PrometheusClient, maxMetricsMetadata = 10000, enricher = defaultEnricher) {
```
juliusv, line 399: Would passing an options struct become more natural at this point, with more and more parameters, all of which are optional? Otherwise if I want to specify an enricher, I also have to specify the first two parameters now.

src/lang-promql/complete/index.ts
```diff
@@ -32,19 +34,20 @@ export interface CompleteStrategy {
 
 // CompleteConfiguration should be used to customize the autocompletion.
 export interface CompleteConfiguration {
-  // Provide these settings when not using a custom PrometheusClient.
-  url?: string;
-  lookbackInterval?: number;
-  httpErrorHandler?: (error: any) => void;
-  fetchFn?: FetchFn;
-
+  remote: {
```
juliusv, line 37: Shouldn't the whole `remote` property also be marked optional?

src/lang-promql/complete/index.ts
```diff
@@ -32,19 +34,20 @@ export interface CompleteStrategy {
...
+    // Provide these settings when not using a custom PrometheusClient.
+    url?: string;
+    lookbackInterval?: number;
+    httpErrorHandler?: (error: any) => void;
+    fetchFn?: FetchFn;
+    // When providing this custom PrometheusClient, the settings above will not be used.
+    prometheusClient?: PrometheusClient;
```
juliusv, line 44: Doesn't have to be in this PR, but would it now make more sense to package the `url` ... `fetchFn` fields into their own `PrometheusConfig`, and then just define the `remote` field as: `remote: PrometheusConfig | PrometheusClient`? Or something like that. Then TypeScript will prevent invalid configuration combinations and things would be clearer.

src/lang-promql/complete/index.ts
Nexucis: ah yeah that is totally true :)

src/lang-promql/complete/index.ts
Nexucis: I won't mind to do it, I would just be sure, it won't be annoying to use. I mean if I want to change just the url, then I have to create the struct:

```typescript
{
  remote: {
    PrometheusConfig: {
      url: 'http://prometheus.land'
    }
  }
}
```
Don't know if it will bother developer to have so much different interface, just to configure an url.
What do you think ?

src/lang-promql/complete/hybrid.ts
Nexucis: mmm yeah that's true. What name do you think we should use for this struct ?

juliusv: `interface HybridCompleteConfig`, and then just `config` here as the parameter name?

src/lang-promql/complete/index.ts
juliusv: The middle step would be left out if you do it like I suggest above:

```typescript
{
  remote: {
    url: 'http://prometheus.land'
  }
}
```

TypeScript would automatically notice that I'm now passing in something that satisfies the interface `PrometheusConfig`, not `PrometheusClient`. Of course then the `url` field should be marked mandatory in `PrometheusConfig`.

Nexucis: ah yeah ok, I see what you meant. Nice thanks, will do that

src/lang-promql/complete/hybrid.ts
Nexucis: mmm actually does it really matter ? In the current design you don't instantiate yourself the HybridComplete. You are passing through the config `CompletionConfig` and through the class `PromQLExtension`.

So after some reflexions, I don't really see the value here. It will create another internal struct / interface

juliusv: Ah yeah, didn't see that. As long as it's only used internally it's fine to keep it like this I think.

README.md
```diff
@@ -66,6 +66,47 @@ In case you would like to deactivate the linter and/or the autocompletion it's s
 const promQL = new PromQLExtension().activateLinter(false).activateCompletion(false) // here the linter and the autocomplete are deactivated
 ```
 
+### maxMetricsMetadata
+
+maxMetricsMetadata is the maximum limit of the number of metrics in Prometheus. Under this limit, it allows the completion to get the metadata of the metrics.
```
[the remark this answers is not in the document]
Nexucis: thanks for your correction !

src/lang-promql/complete/index.ts
Nexucis: done

Nexucis:
@juliusv many thanks for your review. I think I changes everything you asked. Tell me if you are ok know with the changes

juliusv:
:+1: Thanks!

Issue #84, opened 2020-11-18
Extend PrometheusClient interface to pass label matchers as arguments


dsmith3197:
Hi @Nexucis and @juliusv ,

Can we extend the `PrometheusClient` interface to pass in the label matchers for the associated vector selector?  I want to implement a custom `PrometheusClient` that behaves in the same way as the Grafana M3 plugin.  However, the current interface does not pass enough information to make the more refined series queries that the Grafana plugin makes.  This is an issue when a metric name has high enough of a cardinality that will cause the series endpoint to fail when only the metric name is given w/o additional label matchers.  It is also an issue when a user selects a label matcher that then limits the remaining label matchers that can be chosen.  As such, this package as it is now is unusable for my use case.  However, with a slight modification, it will support implementing a `PrometheusClient` with the same functionality as the Grafrana M3 plugin.

A proposed extended interface is as follows:

```
export interface PrometheusClient {
  labelNames(metricName?: string, matchers?: Matcher[]): Promise<string[]>;

  // labelValues return a list of the value associated to the given labelName.
  // In case a metric is provided, then the list of values is then associated to the couple <MetricName, LabelName>
  labelValues(labelName: string, metricName?: string, matchers?: Matcher[]): Promise<string[]>;

  metricMetadata(): Promise<Record<string, MetricMetadata[]>>;

  series(metricName: string, matchers?: Matcher[]): Promise<Map<string, string>[]>;
}
```

```
export interface Context {
  kind: ContextKind;
  metricName?: string;
  labelName?: string;
  matchers?: Matcher[];
}
```

Here are a few examples where this is necessary.  

- Suppose the expression is `some_metric{job="my_job", instance="my_instance", |}` (where the | is the cursor position, not a character in the expression).
  - In this case, codemirror-promql with send the following API request `/api/v1/series?start={}&end={}&matcher[]=some_metric`.  However, I would instead like to make the following API request: `/api/v1/series?start={}&end={}&matcher[]=some_metric{job="my_job", instance="my_instance"}` (this is what the Grafana M3 plugin does).  In order to so, I propose that the `matchers` passed as an argument to `labelValues(...)` is a list of 2 label matchers (i.e., the equivalent of `job="my_job` and `instance="my_instance"`).

- Suppose the expression is `some_metric{job="my_job", instance="|"}` (where the | is the cursor position, not a character in the expression, so the `instance` label matcher currently has an empty string value).
[...]

PR #85, opened 2020-11-20
Add metricNames(prefix?: string) method to prometheus interface

dsmith3197:
This change adds a `metricNames(prefix?: string)` method to the `PrometheusClient` interface and modifies the hybrid autocomplete engine to use this new method when fetching autocomplete results for a metric name.  Importantly, it passes the currently typed metric identifier (i.e., the `prefix`) to the client, which allows for more sophisticated custom clients.

The default client simply calls `labelValues('__name__')` for now, as it did previously.

This change was discussed in https://github.com/prometheus-community/codemirror-promql/issues/84

### DIFF
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -122,15 +122,6 @@ function getMetricNameInVectorSelector(tree: SyntaxNode, state: EditorState): st
   return state.sliceDoc(currentNode.from, currentNode.to);
 }
 
-// getMetricIdentifierValue returns the value of the encapsulating MetricIdentifier for the given tree
-function getMetricIdentifierValue(tree: SyntaxNode, state: EditorState): string {
-  let currentNode: SyntaxNode | null = walkBackward(tree, MetricIdentifier);
-  if (!currentNode) {
-    return '';
-  }
-  return state.sliceDoc(currentNode.from, currentNode.to);
-}
-
 function arrayToCompletionResult(data: Completion[], from: number, to: number, includeSnippet = false, span = true): CompletionResult {
   const options = data;
   if (includeSnippet) {
@@ -260,7 +251,7 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
         // this case is normally impossible since by definition, the identifier has 3 parents,
         // and in Lexer, there is always a default parent in top of everything.
         result.push(
-          { kind: ContextKind.MetricName, metricName: getMetricIdentifierValue(node, state) },
+          { kind: ContextKind.MetricName, metricName: state.sliceDoc(node.from, node.to) },
           { kind: ContextKind.Function },
           { kind: ContextKind.Aggregation }
         );
@@ -272,7 +263,7 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
         if (parent.type.id === BinaryExpr && !containsAtLeastOneChild(parent, 0)) {
           // We are likely in the case 1 or 5
           result.push(
-            { kind: ContextKind.MetricName, metricName: getMetricIdentifierValue(node, state) },
+            { kind: ContextKind.MetricName, metricName: state.sliceDoc(node.from, node.to) },
             { kind: ContextKind.Function },
             { kind: ContextKind.Aggregation },
             { kind: ContextKind.BinOpModifier }
@@ -282,7 +273,7 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
         }
       } else {
         result.push(
-          { kind: ContextKind.MetricName, metricName: getMetricIdentifierValue(node, state) },
+          { kind: ContextKind.MetricName, metricName: state.sliceDoc(node.from, node.to) },
           { kind: ContextKind.Function },
           { kind: ContextKind.Aggregation }
         );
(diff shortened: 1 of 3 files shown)


src/lang-promql/complete/hybrid.ts
```diff
@@ -250,7 +259,11 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
       if (!parent) {
         // this case is normally impossible since by definition, the identifier has 3 parents,
         // and in Lexer, there is always a default parent in top of everything.
-        result.push({ kind: ContextKind.MetricName }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation });
+        result.push(
+          { kind: ContextKind.MetricName, metricName: getMetricIdentifierValue(node, state) },
```
Nexucis, line 263: In the grammar here the node is `Identifier`, so it is really weird that you need to get the node `MetricIdentifier` to get the value. I think you can do it directly with the node `Identifier`.

src/lang-promql/complete/hybrid.ts
```diff
@@ -259,7 +272,7 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
         if (parent.type.id === BinaryExpr && !containsAtLeastOneChild(parent, 0)) {
           // We are likely in the case 1 or 5
           result.push(
-            { kind: ContextKind.MetricName },
+            { kind: ContextKind.MetricName, metricName: getMetricIdentifierValue(node, state) },
```
Nexucis, line 275: same comment

src/lang-promql/complete/hybrid.ts
```diff
@@ -268,7 +281,11 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
           result.push({ kind: ContextKind.BinOp }, { kind: ContextKind.Offset });
         }
       } else {
-        result.push({ kind: ContextKind.MetricName }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation });
+        result.push(
+          { kind: ContextKind.MetricName, metricName: getMetricIdentifierValue(node, state) },
```
Nexucis, line 285: same comment

src/lang-promql/complete/hybrid.ts
```diff
@@ -319,7 +336,11 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
       result.push({ kind: ContextKind.Duration });
       break;
     case FunctionCallBody:
-      result.push({ kind: ContextKind.MetricName }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation });
+      result.push(
+        { kind: ContextKind.MetricName, metricName: getMetricIdentifierValue(node, state) },
```
Nexucis, line 340: I think this case is not really covered by the unit test. Which is a bit sad, since now I don't remember in what case it can happen :(

But here that's not possible that the node `MetricIdentifier` is a parent of the `FunctionCallBody`. 

The grammar is available [here](https://github.com/promlabs/lezer-promql/blob/master/src/promql.grammar)

src/lang-promql/complete/hybrid.ts
Nexucis: so you can simply do `state.sliceDoc(node.from, node.to);`

Nexucis (commented):
thanks for the PR.
I added some comments mostly regarding the way you are getting the metricName.

I would be interested to know when you will need to have the prefix of the metricName, but maybe that's a private knowledge you don't really want to share. Which I can understand.

/cc @juliusv (maybe you will have to watch the repo Julius to avoid to be /cc each time hahaha :p)

src/lang-promql/complete/hybrid.ts
Nexucis: I even wondering what would happen if we remove completely this case. Since the behavior of the lezer-tree changed since the last version, it's possible that it won't be possible to reach this case.

dsmith3197: This case is possible and it is covered by the tests 'metric/function/aggregation autocompletion' and 'metric/function/aggregation autocompletion 3' (I realized this when I removed `metricName: getMetricIdentifierValue(node, state)` from the line above and the tests began to fail because the test expected `metricName` to be the empty string).

dsmith3197:
Hey @Nexucis , thanks for the feedback.  I responded to your comments.

src/lang-promql/complete/hybrid.ts
```diff
@@ -319,7 +327,7 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
       result.push({ kind: ContextKind.Duration });
       break;
     case FunctionCallBody:
-      result.push({ kind: ContextKind.MetricName }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation });
+      result.push({ kind: ContextKind.MetricName, metricName: '' }, { kind: ContextKind.Function }, { kind: ContextKind.Aggregation });
```
Nexucis, line 330: in that case you can remove the metricName.

Nexucis:
thanks @dsmith3197 Just a last a comment and it will be good !

src/lang-promql/complete/hybrid.ts
Nexucis: ah cool. Thanks for the test!

src/lang-promql/complete/hybrid.ts
Nexucis: ah yeah ok no nethermind, it's coherent to do it like that (looking at the test, it would be weird to not having the metricName empty)

Nexucis:
All good @dsmith3197 , thanks for your contribution !

PR #87, opened 2020-11-24
use a LRU caching system

### LINKED ISSUE #71: Limit amount of cached autocomplete data
Currently the Prometheus client never forgets any cached autocomplete data. The amount of series data that can be cached is potentially huge, so keeping all of it forever is basically like a memory leak and can make the browser slow when using the same PromLens tab for a while. It would be good to find a way to expire old items, like an LRU cache method or something like that.

Nexucis:
fix #71

### DIFF
--- a/src/lang-promql/complete/index.ts
+++ b/src/lang-promql/complete/index.ts
@@ -37,6 +37,12 @@ export interface CompleteConfiguration {
   lookbackInterval?: number;
   httpErrorHandler?: (error: any) => void;
   fetchFn?: FetchFn;
+  // cache will allow user to change the configuration of the cached Prometheus client (which is used by default)
+  cache?: {
+    // maxAge is the maximum amount of time that the cache will keep in memory the data fetched from Prometheus.
+    // It is in milliseconds. Default value:  300 000 (5min)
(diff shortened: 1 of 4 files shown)


src/lang-promql/complete/index.ts
```diff
@@ -37,6 +37,12 @@ export interface CompleteConfiguration {
   lookbackInterval?: number;
   httpErrorHandler?: (error: any) => void;
   fetchFn?: FetchFn;
+  // cache will allow user to change the configuration of the cached Prometheus client (which is used by default)
+  cache?: {
+    // maxAge is the maximum amount of time that the cache will keep in memory the data fetched from Prometheus.
```
juliusv, line 42: This is not quite true, as elements > `maxAge` are still kept in memory until you try to retrieve them again. The purpose of `maxAge` seems more to remove outdated / stale elements from the cache so that they are fetched again with an updated version. So a better description for the library user could be:

```suggestion
    // maxAge is the maximum amount of time that a cached completion item is valid before it needs to be refreshed.
```

juliusv:
:+1: Besides comment, thanks!

PR #88, opened 2020-11-30
Support label matchers in Prometheus Client

### LINKED ISSUE #84: Extend PrometheusClient interface to pass label matchers as arguments

Nexucis:
fix #84 

/cc @dsmith3197

### DIFF
--- a/src/lang-promql/parser/matcher.test.ts
+++ b/src/lang-promql/parser/matcher.test.ts
@@ -30,18 +30,21 @@ describe('labelMatchersToString test', () => {
     {
       title: 'metric_name',
       metricName: 'metric_name',
+      labelName: undefined,
       matchers: [] as Matcher[],
       result: 'metric_name',
     },
     {
       title: 'metric_name 2',
       metricName: 'metric_name',
+      labelName: undefined,
       matchers: undefined,
       result: 'metric_name',
     },
     {
       title: 'metric_name{}',
       metricName: 'metric_name',
+      labelName: undefined,
       matchers: [
         {
           type: EqlSingle,
@@ -54,6 +57,7 @@ describe('labelMatchersToString test', () => {
     {
       title: 'sum{LabelName!="LabelValue"}',
       metricName: 'sum',
+      labelName: undefined,
       matchers: [
         {
           type: Neq,
@@ -64,8 +68,9 @@ describe('labelMatchersToString test', () => {
       result: 'sum{LabelName!="LabelValue"}',
     },
     {
-      title: 'rate{labelName=~"label.+"}',
+      title: 'rate{LabelName=~"label.+"}',
       metricName: 'rate',
+      labelName: undefined,
       matchers: [
         {
           type: EqlSingle,
@@ -83,6 +88,7 @@ describe('labelMatchersToString test', () => {
     {
       title: 'rate{LabelName="l1",labelName2=~"label.+",labelName3!~"label.+"}',
       metricName: 'rate',
+      labelName: undefined,
       matchers: [
         {
           type: EqlSingle,
@@ -102,11 +108,39 @@ describe('labelMatchersToString test', () => {
       ] as Matcher[],
       result: 'rate{LabelName="l1",labelName2=~"label.+",labelName3!~"label.+"}',
     },
+    {
+      title: 'rate{LabelName="l1",labelName2=~"label.+",labelName3!~"label.+"}',
+      metricName: 'rate',
+      labelName: 'LabelName',
+      matchers: [
+        {
+          type: EqlSingle,
+          name: 'LabelName',
+          value: 'l1',
+        },
+        {
+          type: EqlRegex,
+          name: 'labelName2',
+          value: 'label.+',
+        },
+        {
+          type: NeqRegex,
+          name: 'labelName3',
+          value: 'label.+',
+        },
+        {
+          type: Neq,
+          name: 'labelName4',
+          value: '',
+        },
+      ] as Matcher[],
+      result: 'rate{labelName2=~"label.+",labelName3!~"label.+"}',
+    },
   ];
 
   testCases.forEach((value) => {
     it(value.title, () => {
-      chai.expect(labelMatchersToString(value.metricName, value.matchers)).to.equal(value.result);
+      chai.expect(labelMatchersToString(value.metricName, value.matchers, value.labelName)).to.equal(value.result);
     });
   });
 });
--- a/src/lang-promql/parser/matcher.test.ts
+++ b/src/lang-promql/parser/matcher.test.ts
@@ -1,0 +1,112 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2020 The Prometheus Authors
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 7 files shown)


src/lang-promql/parser/matcher.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      result: 'metric_name',
+    },
+    {
+      title: 'metric_name 2',
+      metricName: 'metric_name',
+      matchers: undefined,
+      result: 'metric_name',
+    },
+    {
+      title: 'metric_name{}',
```
dsmith3197, line 43: I think this title is out of sync with the test data

src/lang-promql/parser/matcher.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        {
+          type: Neq,
+          name: 'LabelName',
+          value: 'LabelValue',
+        },
+      ] as Matcher[],
+      result: 'sum{LabelName!="LabelValue"}',
+    },
+    {
+      title: 'rate{labelName=~"label.+"}',
```
dsmith3197, line 67: This one also looks like it is out of sync

src/lang-promql/parser/matcher.ts
```diff
@@ -80,3 +58,40 @@ export function buildLabelMatchers(labelMatchers: SyntaxNode[], state: EditorSta
...
 }
+
+export function labelMatchersToString(metricName: string, matchers?: Matcher[]): string {
+  if (!matchers || matchers.length === 0) {
+    return metricName;
+  }
+
+  let matchersAsString = '';
+  for (const matcher of matchers) {
+    if (matcher.value === '') {
```
dsmith3197, line 69: This looks great 😄  One thing, because we have the label name we are trying to match, would it be safer to pass that value into this function and make this `if (matcher.name === labelName || matcher.value === '')`.  If the user initiates autocomplete manually by using the keyboard shortcut, the current way of checking if the value is the empty string would not work as expected (if they have already typed something in the parentheses).

dsmith3197 (changes requested):
This looks great, I left a few small comments 😄  Thank you for doing this.

src/lang-promql/parser/matcher.test.ts
Nexucis: mmm this one I don't get why it's desynch. Can you explain why do you think that please ?

src/lang-promql/parser/matcher.test.ts
Nexucis: oh yeah, good catch, thanks !

src/lang-promql/parser/matcher.ts
Nexucis: ah yeah indeed. Thanks I will change the logic

Nexucis: mmm actually that's a different flow. Here I only did it for the autocompletion of the labelValue. Not sure it makes sense to add the matcher in the labelName prometheus method

src/lang-promql/parser/matcher.test.ts
dsmith3197: If I'm reading the metricName and matchers correctly, the title of this one should be `metric_name{LabelName=''}`.

src/lang-promql/parser/matcher.test.ts
```diff
@@ -0,0 +1,112 @@
...
+          name: 'LabelName',
+          value: '',
+        },
+        {
+          type: EqlRegex,
+          name: 'LabelName',
+          value: 'label.+',
+        },
+      ] as Matcher[],
+      result: 'rate{LabelName=~"label.+"}',
```
dsmith3197, line 81: Also, the result here should include both of the `labelName` matchers right? It's valid to do something like `rate{LabelName="label1", labelName="label1"}`

src/lang-promql/parser/matcher.ts
dsmith3197: Yeah, I agree that it doesn't make sense to add this for the labelName method :) An example of what I was saying above is the user types `rate{labelName="lab|"}` and then manually initiates autocomplete.  In this case, matcher.value will not be the empty string, meaning that we will incorrectly include the `labelName` matcher in the request to /series.  However, if we check that `matcher.name === labelName` we would still correctly remove it from the results vector selector.

src/lang-promql/parser/matcher.test.ts
dsmith3197: wait sorry I realized it's being removed because it's an empty string :)

src/lang-promql/parser/matcher.test.ts
Nexucis: Ah yeah ok. Well in my mind the title was more linked to the result. And since the empty string is removed, well you have at the end an empty bracket

src/lang-promql/parser/matcher.ts
Nexucis: Ah ok. Thanks for the explenation. Makes sense now :). I will make the changes

src/lang-promql/parser/matcher.test.ts
dsmith3197: ahh makes sense :)

Nexucis:
should be good now @dsmith3197 :)

dsmith3197 (approved):
Looks great 😄

Nexucis:
@juliusv I will merge the PR, feel free to yell once you are passing here if you are not agree with the changes :).

PR #91, opened 2020-12-08
Autocomplete bool

### LINKED ISSUE #89: Support `bool` in autocompletion
In `foo > bool 1`, the `bool` is not yet autocompleted.

Nexucis:
fix #89


Nexucis:
Some gif to show different behavior of the bool autocompletion or not: 

### autocomplete bool during the write of the binary expression
bool1

### autocomplete bool after writing the binary expression
bool2

### Do not autocomplete bool when it's not a comparison
not_bool

src/lang-promql/complete/hybrid.ts
```diff
@@ -297,6 +299,12 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
             { kind: ContextKind.Aggregation },
             { kind: ContextKind.BinOpModifier }
           );
+          // in  case the BinaryExpr is a comparison, we should autocomplete the `bool` keyword. But only if it is not present.
+          // To know if it is present, we just have to check if the BoolModifier is present as a child of the BinaryExpr.
+          // If it is the case then it means it is not present.
+          if (containsAtLeastOneChild(parent, Eql, Gte, Gtr, Lte, Lss, Neq) && containsChild(parent, BoolModifier)) {
```
juliusv, line 305: It seems to work great, but could you clarify why bool being present means that it is *not* present? That seems like a contradiction to me without further explanation :)

Nexucis: I updated the comments to explain a bit furthermore why it is like that.
I believe the `BoolModifier` is a direct child of the `BinaryExpr` because of this particular rule:
https://github.com/promlabs/lezer-promql/blob/master/src/promql.grammar#L75

juliusv: Strange! But :+1: :)

PR #97, opened 2021-01-10
upgrade codemirror.next to v0.17.1

Nexucis:


Nexucis:
no big changes so far, it's just a big upgrade of the different dependancies @juliusv. But I think it will have to be released as soon as possible since you cannot use the promQL extension with the version v0.17.x of codemirror.next

juliusv:
:+1: Thank you :)

PR #103, opened 2021-03-05
Support new PromQL func clamp, last_over_time, sgn

Nexucis:
Linked to the modification of PromQL through prometheus/prometheus#8457 and https://github.com/promlabs/lezer-promql/pull/22


juliusv:
:+1: Awesome, thanks :)

PR #104, opened 2021-03-09
Supporting `@` Expr

Nexucis:
This PR is introducing some support regarding the experimental feature `@` in PromQL.:

* After having typed `@`, the system let you autocomplete `start()` and `end()`.
* An highlight of the different keyword is also introduced
* A short linter check is also introduced

Below you can find some screenshot showing the result

image

image

Note: it also contains the upgrade of **typescrit 4**

These changes are linked to these PRs:
* https://github.com/prometheus/prometheus/pull/8425
* https://github.com/prometheus/prometheus/pull/8121
* https://github.com/promlabs/lezer-promql/pull/23

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,10283 @@
 {
   "name": "codemirror-promql",
   "version": "0.12.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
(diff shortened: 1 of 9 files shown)


juliusv:
:+1:

PR #109, opened 2021-03-22
update promQL method signature to support null promise

Nexucis:


juliusv:
:partying_face: :+1:

PR #110, opened 2021-03-22
Remove enricher + bump lezer-promql

### LINKED ISSUE #108: Weird behavior when using enricher to autocomplete query history
It appears that the `enricher` (introduced with #82 ) does't fit really well when it is used to autocomplete an history of PromQL expression.

The issue is that when autocompleting the query proposed, it will replace the current metric typed and not the whole expression.
It is completely expected as long as the `enricher` was used with the context used to autocomplete metrics.

It's just it has a wrong user feeling when doing like that:

https://user-images.githubusercontent.com/4548045/111968181-2f4d0a80-8af9-11eb-8ee9-baf8ed802dbe.mp4

Probably `enricher` should support another kind of context that would be used when you want to replace the whole expression and not just a part of it. 

No idea if it's the best idea to handle that.

Nexucis:
that should close #108 as a "won't fix"

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,38 +102,6 @@ Use it cautiously. A high value of this limit can cause a crash of your browser 
 const promQL = new PromQLExtension().setComplete({ maxMetricsMetadata: 10000 })
 ```
 
-### enricher
-
-enricher is a function that will allow user to enrich the current completion by adding a custom one. It takes two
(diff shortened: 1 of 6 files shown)


juliusv:
:+1: Nice! You're also still going to fix the signatures, right?

PR #120, opened 2021-03-26
add more default snippet that can help to use the aggregation topk, bottomk and count_values

### LINKED ISSUE #118: Add the parameters of the function indescription
When using `topk` or `bottomk`, there is no snippet. The help text does not show the order of those arguments.

We could either add the parameters in the tooltip or have more snippets.

Nexucis:
fix #118 

### DIFF
--- a/src/lang-promql/complete/promql.terms.ts
+++ b/src/lang-promql/complete/promql.terms.ts
@@ -467,4 +467,25 @@ export const snippets: readonly Completion[] = [
     info: 'Set or replace a label value in an input vector',
     apply: snippet('label_replace(${__input_vector__}, "${__dst__}", "${__replacement__}", "${__src__}", "${__regex__}")'),
   },
+  {
+    label: 'topk(__rank_number__, __input_vector__)',
+    type: 'function',
+    detail: 'snippet',
+    info: 'Largest k elements by sample value',
+    apply: snippet('topk(${__rank_number__}, ${__input_vector__})'),
+  },
+  {
+    label: 'bottomk(__rank_number__, __input_vector__)',
+    type: 'function',
+    detail: 'snippet',
+    info: 'Smallest k elements by sample value',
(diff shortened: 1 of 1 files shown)


src/lang-promql/complete/promql.terms.ts
```diff
@@ -467,4 +467,25 @@ export const snippets: readonly Completion[] = [
     info: 'Set or replace a label value in an input vector',
     apply: snippet('label_replace(${__input_vector__}, "${__dst__}", "${__replacement__}", "${__src__}", "${__regex__}")'),
   },
+  {
+    label: 'topk(__top_rank_number__, __metric__)',
```
juliusv, line 471: Maybe just `__rank__` or `__rank_number__`? Is rank the right term for something like topk/bottomk? But at least the `top`/`bottom` seems a bit redundant.

`__metric__` means either a single time series or a whole group of series with the same metric name, neither of which is precisely the input requirement here. I've used `__input_vector__` for some of the other snippets, maybe keep that in the new snippets too?

Nexucis: sounds good to me. I wasn't really confortable with what I put here anyway :). Thanks for your help !

src/lang-promql/complete/promql.terms.ts
```diff
@@ -467,4 +467,25 @@ export const snippets: readonly Completion[] = [
...
+    type: 'function',
+    detail: 'snippet',
+    info: 'Smallest k elements by sample value',
+    apply: snippet('bottomk(${__bottom_rank_number__}, ${__metric__})'),
+  },
+  {
+    label: 'count_values("__label_name__", __metric__)',
+    type: 'function',
+    detail: 'snippet',
+    info: 'Count number of elements with the same value',
```
juliusv, line 488: To make it even clearer:

```suggestion
    info: 'Count the number of series per distinct sample value',
```

juliusv:
:+1:

PR #125, opened 2021-04-02
Fix the autocompletion of the label after a comma

### LINKED ISSUE #123: PromQL autocomplete: impossible to get full labels list after comma
Hello

I found a bug in the prometheus UI implementation,
When I start to type a promQL, after brackets, the full list is proposed automatically (or by hitting ctrl+space), but this is not the case starting second label filtering, the labels autocompletion is not proposed.

I tried to record this in the GIF below, at the very end I am trying to hit ctrl+space as well without success...
I would expect the labels completion dropdown

Apr-02-2021 15-59-46

Nexucis:
fix #123 

label

+ simplify a bit the development section as a side effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,17 +175,6 @@ npm start
 
 This should create a tab in your browser with the development app that contains CodeMirror Next with the PromQL plugin.
 
-### Autocompletion
-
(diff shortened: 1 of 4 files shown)


juliusv:
:+1: Nice, thanks!

PR #126, opened 2021-04-06
remove angular example in favor of live example from prometheus and perses

Nexucis:
The preview will be replaced by the URL from Netifly. Like that also we are sure to have an up to date preview


README.md
```diff
@@ -41,7 +41,7 @@ npm install --save @codemirror/basic-setup
 
 You can try out the latest release version of this mode on:
```
juliusv, line 42: This isn't true anymore because it's deployed from master and not the latest release.

juliusv:
:+1: besides comment

README.md
Nexucis: thanks for the review, I updated this line :). Will merge the PR.

PR #127, opened 2021-04-07
provide getter and setter to manipulate the different object provided by the lib

### LINKED ISSUE #124: Have a way to warm the cache
When the autocompletion is contacting a huge Thanos, then when you are starting to type a PromQL expression, nothing is proposed until the request to get a list of metric is returned.

It's a bit annoyed since it is blocking the autocompletion of the PromQL keyword such as the function/aggregation.

A way to avoid that would be to warm the cache embedded into the PrometheusClient used by the lib by performing the query to get the list of the metric before it is actually using.

The problem is there is no way to get the instance of the PrometheusClient used by lib.

1. First option is to simply provide a getter that would return the PrometheusClient instance used. But I don't really want to change the interface `CompleteStrategy`. Having a PrometheusClient is more related to the implementation of the interface than to the interface itself.
2. Another option would be to provide a option `cache.warm`, if set, then it will warm the cache when the PromQLExtension is instanciated. But I don't have a really good feeling ether for this solution.

Nexucis:
It's a way to fix #124.
The lib should not provide the way to warm a cache, just enough flexibility if someone would like too. And from my point of view, having a way to get the instance of the cached prometheus client should be good. Because then to "warm" it, you just need to get the instance of the prometheusClient and then manually perform the retrieve of the metric list for example

### DIFF
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -444,6 +444,10 @@ export class HybridComplete implements CompleteStrategy {
     this.maxMetricsMetadata = maxMetricsMetadata;
   }
 
+  getPrometheusClient(): PrometheusClient | undefined {
+    return this.prometheusClient;
+  }
+
   promQL(context: CompletionContext): Promise<CompletionResult | null> | CompletionResult | null {
     const { state, pos } = context;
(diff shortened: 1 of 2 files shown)


juliusv:
:+1: Thanks, seems fine enough to me for now :)

PR #130, opened 2021-04-09
Add CommonJS build

dsmith3197:
This adds a CommonJS module to the final build.  The build folder now has the following structure
```
build/
  cjs/
    ...
  esm/
    ...
```
I moved the ES2105 module from the root directory into `esm/` so that the two modules don't overlap.  This is a small breaking change with respect to imports from files other than the index file, so I can change it back if this is not desired.

To fix the tests, I added `ts-mocha` and turned on `esModuleInterop`.


Nexucis:
really cool, thanks :).

> This is a small breaking change with respect to imports from files other than the index file, so I can change it back if this is not desired.

I'm not sure I understood what you mean.
Does it mean that when using this lib, the import will have to change ?

For example, from `import { PromQLExtension } from 'codemirror-promql'` to `import { PromQLExtension } from 'codemirror-promql/esm';`  ?

dsmith3197:
> For example, from `import { PromQLExtension } from 'codemirror-promql'` to `import { PromQLExtension } from 'codemirror-promql/esm';` ?

So `import { PromQLExtension } from 'codemirror-promql'` will remain the same because the `main` and `module` fields in package.json point to `cjs/index.js` and `esm/index.js` respectively.  

What will change is an import like `import { labelMatchersToString } from 'codemirror-promql/parser/matcher';`.  It will now be 
`import { labelMatchersToString } from 'codemirror-promql/esm/parser/matcher';` or `import { labelMatchersToString } from 'codemirror-promql/cjs/parser/matcher';` depending upon which version you want to use.

Nexucis:
mmm ok, thanks for the explanation :). And is it possible to merge both module in the same folder ? I think codemirror doesn't have these split right ? 

Just wondering what is the impact if we merge both folder ?

dsmith3197:
The codemirror modules export a single file, which is how they are able to achieve that. I'm not sure if there is a way around this because this package exports multiple files.  Anyways, consumers should probably only be importing things from the default index.js file because that's what is directly supported.

dsmith3197:
Sorry I didn't answer your second question.  We can merge the folders because they have the exact same files but with different content (i.e., one is transpiled to cjs and the other to es2015).

Nexucis:
great thanks for the answers !
I think I would prefer to have everything in the same folder. According to what you said, we can avoid a breaking change like that, and I believe it will ease the way to use it then :).

Nexucis:
Sooo, I took some times to read a bit more about what does it mean to support CommonJS and ESM module at the same time and about the different possible approach. Not because what you did didn't convince, but just because I wasn't sure to understand exactly what you did and why.

And actually , I think you took the best and the simplest way to handle it. So I think I will just merge it like that. No need to change anything.

For others that would like to understand it a bit better, here some blogs that could help:

* https://2ality.com/2019/10/hybrid-npm-packages.html
* https://dev.to/remshams/rolling-up-a-multi-module-system-esm-cjs-compatible-npm-library-with-typescript-and-babel-3gjg

Thank you very much for these changes @dsmith3197. It will help a lot people that would like to use this lib !

PR #139, opened 2021-04-25
fetch series with post method

faceair:
ref: https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers
> You can URL-encode these parameters directly in the request body by using the POST method and Content-Type: application/x-www-form-urlencoded header. This is useful when specifying a large or dynamic number of series selectors that may breach server-side URL character limits.

### DIFF
--- a/src/lang-promql/client/prometheus.ts
+++ b/src/lang-promql/client/prometheus.ts
@@ -182,13 +182,16 @@ export class HTTPPrometheusClient implements PrometheusClient {
   series(metricName: string, matchers?: Matcher[], labelName?: string): Promise<Map<string, string>[]> {
     const end = new Date();
     const start = new Date(end.getTime() - this.lookbackInterval);
-    const params: URLSearchParams = new URLSearchParams({
+    const body: URLSearchParams = new URLSearchParams({
       start: start.toISOString(),
       end: end.toISOString(),
       'match[]': labelMatchersToString(metricName, matchers, labelName),
     });
     // See https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers
-    return this.fetchAPI<Map<string, string>[]>(`${seriesEndpoint}?${params}`).catch((error) => {
+    return this.fetchAPI<Map<string, string>[]>(`${seriesEndpoint}`, {
+      method: 'post',
+      body: body,
+    }).catch((error) => {
       if (this.errorHandler) {
         this.errorHandler(error);
       }
@@ -200,8 +203,8 @@ export class HTTPPrometheusClient implements PrometheusClient {
     return this.labelValues('__name__');
   }
 
-  private fetchAPI<T>(resource: string): Promise<T> {
-    return this.fetchFn(this.url + resource)
+  private fetchAPI<T>(resource: string, init?: RequestInit): Promise<T> {
+    return this.fetchFn(this.url + resource, init)
       .then((res) => {
         if (!res.ok && ![badRequest, unprocessableEntity, serviceUnavailable].includes(res.status)) {
           throw new Error(res.statusText);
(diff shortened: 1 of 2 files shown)


src/lang-promql/client/prometheus.ts
```diff
@@ -188,7 +188,7 @@ export class HTTPPrometheusClient implements PrometheusClient {
       'match[]': labelMatchersToString(metricName, matchers, labelName),
     });
     // See https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers
-    return this.fetchAPI<Map<string, string>[]>(`${seriesEndpoint}?${params}`).catch((error) => {
+    return this.fetchAPI<Map<string, string>[]>(`${seriesEndpoint}?${params}`, {method: 'post'}).catch((error) => {
```
Nexucis, line 191: in case it is really usefull to move to the POST method (looking forward to the answer to my above question :)), then you also have to put the parameter in the body and to set the content type like it is explained in the documentation you pointed.

For the moment, the only effect of this PR is the HTTP method changed from GET to POST. But the query parameters are still there and the content type didn't change.

image

src/test/utils.ts
```diff
@@ -46,6 +46,9 @@ export function mockPrometheusServer() {
     .replyWithFile(200, __dirname + '/metadata.json')
     .get('/api/v1/series')
```
Nexucis, line 47: I guess this, can be removed since it won't be possible to call the endpoint with the method GET

src/test/utils.ts
```diff
@@ -46,6 +46,9 @@ export function mockPrometheusServer() {
...
     .query(true)
+    .replyWithFile(200, __dirname + '/alertmanager_alerts_series.json')
+    .post('/api/v1/series')
+    .query(true)
```
Nexucis, line 51: you can remove this set `query(true)` as you shouldn't have any query parameter

src/lang-promql/client/prometheus.ts
```diff
@@ -200,8 +200,8 @@ export class HTTPPrometheusClient implements PrometheusClient {
     return this.labelValues('__name__');
   }
 
-  private fetchAPI<T>(resource: string): Promise<T> {
-    return this.fetchFn(this.url + resource)
+  private fetchAPI<T>(input: string, init?: RequestInit): Promise<T> {
```
Nexucis, line 203: `resource` was good for me. I don't see the reason to change it to `input`. It's not really clearer from my point of view.

Nexucis (changes requested):
I left few comments in the code. But first please give some context to understand exactly why this PR is needed :).
Thanks !

Nexucis:
Hi,
thanks for the PR.
Before merging this PR I would be glad to have some context of how you reached the limit. The usage of the series endpoint should be quite limited in this lib, so for me it's quite impossible to reach the limit. 

So can you please explain how did you do ?

src/lang-promql/client/prometheus.ts
faceair: image
I just saw that the original parameter name for the fetch api was `input`, so I changed it to `input` here. if you like the name `resource`, we can keep it.

faceair:
Thanks for the review, I am really not good at js.
My colleague told me that he had encountered a case where the url was too long and was rejected by nginx, and I didn't get a full example. I'll try to tweak the way I send requests to avoid this problem.

Nexucis (approved):
looks to me, thanks !

juliusv:
Note that this breaks codemirror-promql usage against proxied Grafana datasources:

I get a `403 Forbidden` with an error message of:

`{"message":"Posts not allowed on proxied Prometheus datasource except on /query and /query_range"}`

Nexucis:
:( Indeed I didn't think at this usage. I created the issue #144 based on your comment, just to be sure the bug won't be lost.

Of course if you want to open a PR to fix it directly, it would cool too :D.

PR #141, opened 2021-05-03
Autocomplete number when needed

### LINKED ISSUE #136: Support autocompleting NaN and Inf
Currently when I start typing `NaN` or `Inf` in positions where they are allowed (anywhere where scalar values are allowed, which could be positions where vectors are also allowed), I only get metric names as autocompletion results. It would be great if we could autocomplete `NaN` and `Inf` as well (capitalized or not).

Nexucis:
fix #136

### DIFF
--- a/src/lang-promql/complete/hybrid.ts
+++ b/src/lang-promql/complete/hybrid.ts
@@ -36,6 +36,7 @@ import {
   EqlRegex,
   EqlSingle,
   Expr,
+  FunctionCallArgs,
   FunctionCallBody,
   GroupingLabel,
   GroupingLabels,
@@ -79,6 +80,7 @@ import {
   durationTerms,
   functionIdentifierTerms,
   matchOpTerms,
+  numberTerms,
   snippets,
 } from './promql.terms';
 import { Matcher } from '../types';
@@ -93,6 +95,7 @@ const autocompleteNodes: { [key: string]: Completion[] } = {
   functionIdentifier: functionIdentifierTerms,
   aggregateOp: aggregateOpTerms,
   aggregateOpModifier: aggregateOpModifierTerms,
+  number: numberTerms,
 };
 
 // ContextKind is the different possible value determinate by the autocompletion
@@ -112,6 +115,7 @@ export enum ContextKind {
   Offset,
   Bool,
   AtModifiers,
+  Number,
 }
 
 export interface Context {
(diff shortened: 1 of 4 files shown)


src/lang-promql/complete/promql.terms.ts
```diff
@@ -445,6 +445,11 @@ export const aggregateOpModifierTerms = [
   },
 ];
 
+export const numberTerms = [
+  { label: 'nan', info: 'not a number', type: 'constant' },
```
juliusv, line 449: I'd go with:

```suggestion
  { label: 'NaN', info: 'Floating-point NaN value', type: 'constant' },
```

src/lang-promql/complete/promql.terms.ts
```diff
@@ -445,6 +445,11 @@ export const aggregateOpModifierTerms = [
...
+  { label: 'inf', info: 'the infinite and beyond of the number', type: 'constant' },
```
juliusv, line 450: I'd go with:

```suggestion
  { label: 'Inf', info: 'Floating-point infinity', type: 'constant' },
```

src/lang-promql/complete/hybrid.ts
```diff
@@ -330,12 +336,22 @@ export function analyzeCompletion(state: EditorState, node: SyntaxNode): Context
           { kind: ContextKind.Function },
           { kind: ContextKind.Aggregation }
         );
+        if (parent.type.id !== FunctionCallArgs && parent.type.id !== MatrixSelector) {
```
juliusv, line 339: Hmm, when I type `n` in a function call parameter, it now doesn't autocomplete `nan`, but should. Is it because of this?

Nexucis: ah yeah that's possible. It was more because we don't want to autocomplete the number in the the `sum` for example.
But yeah I forgot that sometimes it's possible ....

Maybe we should just autocomplete it everytime even if it's sometimes false depending of the aggregation / function

Nexucis: even if I don't really like it :/.

But well in the other side, putting NaN / Inf is a bit a corner case in the function call parameters right ?

juliusv: Yeah, it's legal in the type system and will cause a valid query, but I'm not aware of any valid use case with the current set of functions taking scalars :) So probably it's fine to just not support it for now, yeah.

src/lang-promql/complete/promql.terms.ts
Nexucis: fixed

src/lang-promql/complete/promql.terms.ts
Nexucis: fixed

juliusv:
:+1:

PR #142, opened 2021-05-19
Integrate lezer-promql v0.19.0 with alternative top

Nexucis:


src/app/app.html
```diff
@@ -34,6 +34,12 @@ <h3>CodeMirror Mode PromQL</h3>
   <option selected value="offline">Offline</option>
   <option value="prometheus">Prometheus</option>
 </select>
+<br>
+<label for="fullLanguage">choose if you want to use the complete PromQL language</label>
+<select name="fullPromQLLanguage" id="fullLanguage">
```
juliusv, line 39: How about something like

```
Language to complete:
* Full PromQL
* Metric names
```

...because otherwise it's unclear what "no" means.

Nexucis: sounds good

src/lang-promql/promql.ts
```diff
@@ -102,10 +105,12 @@ export class PromQLExtension {
     return this;
   }
 
-  asExtension(): Extension {
-    let extension: Extension = [promQLLanguage];
+  asExtension(fullPromQLLanguage = true): Extension {
```
juliusv, line 108: I would avoid a boolean option here, also because we may want to add more sub-types (label names, label values). Maybe a `LanguageType` enum with values `PromQL` and `MetricName` instead?

Nexucis: mmm perhaps later. It would require more thought to integrate it properly and for the moment it doesn't worth it I think (I mean to manage the LanguageType in the complete / linter)

src/app/app.html
```diff
@@ -34,6 +34,12 @@ <h3>CodeMirror Mode PromQL</h3>
   <option selected value="offline">Offline</option>
   <option value="prometheus">Prometheus</option>
 </select>
+<br>
+<label for="fullLanguage">Language to complete</label>
```
juliusv, line 38: I guess this and the line below now also shouldn't say `fullLanguage` / `fullPromQLLanguage` anymore, but something like `languageType`.

juliusv:
:+1: besides comment

PR #146, opened 2021-06-25
be able to configure the http method used

### LINKED ISSUE #144: fetch series with POST not allowed by Grafana
Note that this breaks codemirror-promql usage against proxied Grafana datasources:

I get a `403 Forbidden` with an error message of:

`{"message":"Posts not allowed on proxied Prometheus datasource except on /query and /query_range"}`

_Originally posted by @juliusv in https://github.com/prometheus-community/codemirror-promql/issues/139#issuecomment-848716101_

Nexucis:

fix #144

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,8 +142,8 @@ const promQL = new PromQLExtension().setComplete({ remote: { fetchFn: myHTTPClie
 
 ###### Duration to use for looking back when retrieving metrics / labels
 
-If you are a bit familiar with the Prometheus API, you are aware that you can pass a time interval that is used to
-tell to Prometheus which period of time you are looking for when retrieving metadata (like metrics / labels).
+If you are a bit familiar with the Prometheus API, you are aware that you can pass a time interval that is used to tell
+to Prometheus which period of time you are looking for when retrieving metadata (like metrics / labels).
 
 In case you would like to provide your own duration, you can override the variable `lookbackInterval`. By default, the
 value is `12 * 60 * 60 * 1000` (12h). The value must be defined in **milliseconds**.
@@ -159,6 +159,17 @@ Prometheus.
 
 ```typescript
 const promQL = new PromQLExtension().setComplete({ remote: { httpErrorHandler: (error: any) => console.error(error) } })
+```
+
+###### HTTP method used
+
+By default, the Prometheus client will use the HTTP method `POST` when contacting Prometheus for the
+endpoints `/api/v1/labels` and `/api/v1/series`.
+
+You can change it to use the HTTP method `GET` if you prefer.
+
+```typescript
+const promQL = new PromQLExtension().setComplete({ remote: { httpMethod: 'GET' } })
 ```
 
 ###### Cache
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -161,6 +161,17 @@ Prometheus.
...
 
+###### HTTP method used
+
+By default, the prometheus client will use the HTTP method `POST` when contacting Prometheus for the
+endpoints `/api/v1/labels` and `/api/v1/series`.
+
+You can change it to use the HTTP method `GET` if you prefer.
+
+```typescript
+const promQL = new PromQLExtension().setComplete({ remote: { httpMethodUsed: 'get' } })
```
juliusv, line 172: I feel `httpMethodUsed` is unnecessarily verbose. Better IMO would be just `method` or `httpMethod`.

README.md
```diff
@@ -177,24 +188,24 @@ const promQL = new PromQLExtension().setComplete({ remote: { cache: { maxAge: 5
 
 ###### Initial Metric List
 
-The cache can be initialized with a list of metrics name. It is useful when you already have the list of
-the metrics somewhere else in your application, and you would like to share this list with the embedded Prometheus
-client of `codemirror-promql`.
+The cache can be initialized with a list of metrics name. It is useful when you already have the list of the metrics
```
juliusv, line 191:
```suggestion
The cache can be initialized with a list of metric names. It is useful when you already have the list of the metrics
```

README.md
```diff
@@ -161,6 +161,17 @@ Prometheus.
 const promQL = new PromQLExtension().setComplete({ remote: { httpErrorHandler: (error: any) => console.error(error) } })
 ```
 
+###### HTTP method used
+
+By default, the prometheus client will use the HTTP method `POST` when contacting Prometheus for the
```
juliusv, line 166:
```suggestion
By default, the Prometheus client will use the HTTP method `POST` when contacting Prometheus for the
```

README.md
```diff
@@ -177,24 +188,24 @@ const promQL = new PromQLExtension().setComplete({ remote: { cache: { maxAge: 5
...
-    cache: {
-      initialMetricList: ['ALERTS',
-        'ALERTS_FOR_STATE',
-        'alertmanager_alerts',
-        'alertmanager_alerts_invalid_total',
-        'alertmanager_alerts_received_total',
-      ]
+    remote: {
+        cache: {
+            initialMetricList: ['ALERTS',
```
juliusv, line 201: Wouldn't it be nicer to have `'ALERTS',` on its own line as well?

src/lang-promql/client/prometheus.ts
```diff
@@ -68,6 +68,7 @@ export interface PrometheusConfig {
   fetchFn?: FetchFn;
   // cache will allow user to change the configuration of the cached Prometheus client (which is used by default)
   cache?: CacheConfig;
+  httpMethodUsed?: 'post' | 'get';
```
juliusv, line 71: Would it be better to standardize on the upper-cased `POST` and `GET`, as the official examples in https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch and the fetch spec (https://fetch.spec.whatwg.org/#methods) use?

src/lang-promql/client/prometheus.ts
```diff
@@ -101,18 +103,27 @@ export class HTTPPrometheusClient implements PrometheusClient {
...
 
   labelNames(metricName?: string): Promise<string[]> {
     const end = new Date();
     const start = new Date(end.getTime() - this.lookbackInterval);
     if (metricName === undefined || metricName === '') {
-      const params: URLSearchParams = new URLSearchParams({
-        start: start.toISOString(),
-        end: end.toISOString(),
-      });
+      const params = this.buildURIAndBody(
```
juliusv, line 115: At this point, `params` does not seem like the right name anymore, since it stores more than just the URL params. How about `request` or something like that, in which case `buildURIAndBody()` could be called `buildRequest()` as well?

src/lang-promql/client/prometheus.ts
Nexucis: yeah totally make sense. I don't even know why I didn't put it already in uppercase. It's a bit a common definition across the different language.

src/lang-promql/client/prometheus.ts
Nexucis: I didn't feel confortable with this name anyway ^^. `request` + `buildRequest()` are fine for me

Nexucis:
should be good like that @juliusv.

README.md
```diff
@@ -177,24 +188,25 @@ const promQL = new PromQLExtension().setComplete({ remote: { cache: { maxAge: 5
 
 ###### Initial Metric List
 
-The cache can be initialized with a list of metrics name. It is useful when you already have the list of
-the metrics somewhere else in your application, and you would like to share this list with the embedded Prometheus
-client of `codemirror-promql`.
+The cache can be initialized with a list of metrics names. It is useful when you already have the list of the metrics
```
juliusv, line 191:
```suggestion
The cache can be initialized with a list of metric names. It is useful when you already have the list of the metrics
```

juliusv:
:+1: After last nit is fixed :)

PR #151, opened 2021-09-25

prombot:


Nexucis:
@roidelapluie it's almost good ! Just it shouldn't remove the circleci config, the licence and other .md files.
Perhaps an easy way for the circleci config is just to add it back in prometheus.
But perhaps you prefer a smarter way ?

Nexucis:
@juliusv: can you remove the netifly config. The app package doesn't existe anymore so netifly is in error

juliusv:
Deleted it! I think we just have to ignore the Netlify checks on this PR here then.

roidelapluie:
> @roidelapluie it's almost good ! Just it shouldn't remove the circleci config, the licence and other .md files.
> Perhaps an easy way for the circleci config is just to add it back in prometheus.
> But perhaps you prefer a smarter way ?

There is a smarter way. But I think we need to change the script. It is using git restore. Maybe the git version does not support it and we have to get back to git checkout.

roidelapluie:
git: 'restore' is not a git command. See 'git --help'.

Nexucis:
mmm probably we could just upgrade git don't you think @roidelapluie ?

roidelapluie:
https://github.com/prometheus/golang-builder/pull/120 not-as-easy-as-it-looks

Nexucis:
I will take a look at this PR then

PR #153, opened 2021-09-26

prombot:


Nexucis:
@roidelapluie it works ! this is so cool \o/.

Last thing would be to re-enable the first sync for the license. I think @SuperQ did something to avoid to change the current license. (It can explain why the license didn't change from MIT to Apache v2)

roidelapluie:
You need to push the new licence manually in a pull request.

